PR #49, opened 2023-11-22
remove legacy APIs

siyangqiu:
Remove deprecated OpenAI APIs


areibman:
We should close this on January 4th, 2024
"Most models that support the legacy Completions endpoint [will be shut off on January 4th, 2024](https://platform.openai.com/docs/deprecations/2023-07-06-gpt-and-embeddings)."

https://platform.openai.com/docs/api-reference/completions

areibman (approved):
LGTM

PR #84, opened 2024-02-09
83 update events prompts to be sent as chatml

### LINKED ISSUE #83: Update events prompts to be sent as ChatML
## 🚀 Feature Request

Currently, we do not consistently send prompts to the server as either ChatML format or string. We should enforce this standard on SDK for both langchain and OpenAI

areibman:


siyangqiu:
Implemented in v0.1

PR #85, opened 2024-02-21
Refactor new db structure

braelyn-ai:
## 📥 Pull Request

**📘 Description**
Moves the repo officially into the v0.1 phase.
Legacy support and archives will be on the v0.0 branch.


siyangqiu (approved):
LGTM - just maybe not call reference the new db structure because that work is not nearly complete

PR #89, opened 2024-02-28
feature: client support starting new sessions

siyangqiu:
## 📥 Pull Request

**📘 Description**

*Cleanup the constructor to be clearer by moving the exit handling code into a function
*End_session sets session and worker to None, thereby allowing new sessions to be created
*Added checks to start_session so that session will not start unless necessary information is provided
*Cleaned up some error messages
*use new api endpoint at api.agentops.ai

**🧪 Testing**
tox

and
```
from agentops import Client, Event

ao_client = Client(auto_start_session=False, tags=['test'])
ao_client.start_session()
ao_client.record(Event(event_type="test2"))
ao_client.end_session('Success')

ao_client.start_session()
ao_client.start_session()
ao_client.record(Event(event_type="test2"))
ao_client.end_session('Success')
```
We get a warning `WARNING:root:AgentOps: Cannot start session - Session already started` on the session that was started without an end_sesssion.


braelyn-ai (approved):

agentops/client.py
```diff
@@ -148,10 +104,9 @@ def record(self, event: Event):
             self._worker.add_event(
                 {'session_id': self._session.session_id, **event.__dict__})
         else:
-            logging.info("This event was not recorded because the previous session has been ended" +
-                         " Start a new session to record again.")
+            logging.warn("AgentOps: Cannot record event - no current session")
 
-    def record_action(self, event_name: str, tags: Optional[List[str]] = None):
+    def record_function(self, event_name: str, tags: Optional[List[str]] = None):
```
braelyn-ai, line 109: Why `record_function`?

siyangqiu: We're using 'action' as an event type and record action was never really that great of a name. I'm open to other names

PR #91, opened 2024-02-29
Multi-Agent Support

braelyn-ai:
## 📥 Pull Request

- Creates a new Agent class that wraps an LLM completion function.
- Create an agent with `agent1 = client.create_agent()`.
- All calls with the agent will be tracked and tied to that agent.
- Adds an example notebook
- Moves `safe_serialize` to `helpers.py`


examples/multi-agent.ipynb
areibman, line 1: image
Please re-run the notebook. First cell should be `[1]`. Also, there is a giant error

areibman, line 1: Nit: This is the more pythonic way to do this:
`OPENAI_API_KEY = os.environ.get('OPENAI_API_KEY', "<your_api_key")`
image
The second arg will be taken if the variable is not found

areibman, line 1: image
* Remove endpoint arg. Users don't need to see this
* You're probably using openai version 0.2x.xx. Upgrade to 1.x.x+

agentops/worker.py
areibman, line 1: image
We're moving towards the Otel way of tracking agents. We don't want to be in the business of creating our own framework, so let's not do this

agentops/agent.py
```diff
@@ -0,0 +1,139 @@
...
+        # response = Completions.create(*args, **kwargs)
+
+        # if the response is a generator, decorate the generator
+        if isinstance(response, Stream):
+            def generator():
+                for chunk in response:
+                    self._handle_stream_chunk(chunk, init_timestamp, **kwargs)
+                    yield chunk
+
+            return generator()
```
areibman, line 55: This is copied from `llm_tracker.py`, no? Why not update that instrumentor instead?

agentops/agent.py
```diff
@@ -0,0 +1,139 @@
+from .event import Event
+from .http import HttpClient
+from .helpers import safe_serialize, get_ISO_time, Providers
+from openai import Stream, AsyncStream
+from openai.types.chat import ChatCompletionChunk
+from openai.resources import AsyncCompletions
+import openai
+
+
+class Agent:
```
areibman, line 10: We need to have a longer discussion about the implementation pattern. I'd prefer to not have strongly opinionated agent classes if we can help it-- we should try to rely on 3rd party instrumentors and lightweight decorators/context managers (`with`)

areibman (changes requested):

braelyn-ai:
Abandoned in favor of a better design

PR #97, opened 2024-03-01
moves session_id out of event and to batched request

braelyn-ai:
## 📥 Pull Request

Events that are batched will always share the same `session_id`. This changes the id to send outside of the sessions.

### DIFF
--- a/agentops/worker.py
+++ b/agentops/worker.py
@@ -5,6 +5,7 @@ from .http import HttpClient
 from .config import Configuration
 from .session import Session
 from typing import Dict
+import logging
 
 
 def is_jsonable(x):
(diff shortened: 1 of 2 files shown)


agentops/worker.py
```diff
@@ -47,6 +48,7 @@ def __init__(self, config: Configuration) -> None:
         self.thread = threading.Thread(target=self.run)
         self.thread.daemon = True
         self.thread.start()
+        self._session: Session | None = None
```
siyangqiu, line 51: Are there considerations for pulling in just the session_id or the whole thing? Not a big deal either way.

siyangqiu (approved):
Looks good to me. Feel free to ignore the comment.

PR #102, opened 2024-03-08
Multi agent support

braelyn-ai:
## 📥 Pull Request

- Creates Agent decorator
  - Can be used to assign classes or functions as agents
- Using the decorator creates an agent in the db
- Singleton decorator
- Makes `Client` adhere to a singleton pattern
- Jupyter Notebook example

Small changes:
- Removes frequency recording in env as it does not work on M1 chip macs
- `warn` -> `warning`
- `print` -> `logging`

### DIFF
--- a/agentops/llm_tracker.py
+++ b/agentops/llm_tracker.py
@@ -135,7 +135,7 @@ class LlmTracker:
                 ))
                 # Standard response
             except Exception as e:
-                print(
+                logging.error(
                     f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
 
         return response
@@ -187,7 +187,7 @@ class LlmTracker:
                     self.client.record(self.event_stream)
                     self.event_stream = None
             except Exception as e:
-                print(
+                logging.error(
                     f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
 
         # if the response is a generator, decorate the generator
@@ -234,7 +234,7 @@ class LlmTracker:
             ))
             # Standard response
         except Exception as e:
-            print(
+            logging.error(
                 f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
 
         return response
--- a/agentops/llm_tracker.py
+++ b/agentops/llm_tracker.py
@@ -9,17 +9,15 @@ from .helpers import get_ISO_time
 
 
 def check_call_stack_for_agent_id() -> str | None:
-    # only looks at the last 2 frames otherwise it will retrieve
-    # whatever agent class was used last
     for frame_info in inspect.stack():
-        # Get the locals from the current frame
+        # Look through the call stack for the class that called the LLM
         local_vars = frame_info.frame.f_locals
         for var in local_vars.values():
+            # We stop looking up the stack at main because after that we see global variables
             if var == "__main__":
                 return
-            # Instead of checking if var is a class, check if it's an instance with the _is_ao_agent attribute
             if hasattr(var, '_is_ao_agent') and getattr(var, '_is_ao_agent'):
-                logging.info('LLM call from agent named: ' + getattr(var, '_ao_agent_name'))
+                logging.debug('LLM call from agent named: ' + getattr(var, '_ao_agent_name'))
                 return getattr(var, '_ao_agent_id')
     return None
 
@@ -78,8 +76,8 @@ class LlmTracker:
                     self.client.record(self.event_stream)
                     self.event_stream = None
             except Exception as e:
-                logging.info(
-                    f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
+                logging.warning(
+                    f"AgentOps Error: Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
 
         # if the response is a generator, decorate the generator
         if inspect.isasyncgen(response):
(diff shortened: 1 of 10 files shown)


agentops/llm_tracker.py
```diff
@@ -3,10 +3,27 @@
...
+        # Get the locals from the current frame
+        local_vars = frame_info.frame.f_locals
+        for var in local_vars.values():
+            if var == "__main__":
+                return
+            # Instead of checking if var is a class, check if it's an instance with the _is_ao_agent attribute
+            if hasattr(var, '_is_ao_agent') and getattr(var, '_is_ao_agent'):
+                logging.info('LLM call from agent named: ' + getattr(var, '_ao_agent_name'))
+                return getattr(var, '_ao_agent_id')
+    return None
```
siyangqiu: The comments are outdated <-- which is why I typically don't like them. But in this case, I think we should leave an explanation as to what this whole thing is for and another comment explaining that the `var == __main__` piece.

agentops/llm_tracker.py
```diff
@@ -59,7 +78,7 @@ def handle_stream_chunk(chunk):
                     self.client.record(self.event_stream)
                     self.event_stream = None
             except Exception as e:
-                print(
+                logging.info(
```
siyangqiu: I have all of the my logs start with "AgentOps:". I think we should try to be consistent. Also, I think this deserves at least a logging.warning().

agentops/llm_tracker.py
```diff
@@ -115,7 +135,7 @@ def generator():
                 ))
                 # Standard response
             except Exception as e:
-                print(
+                logging.error(
                     f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
```
siyangqiu: same as above, except the logging level isn't consistent

agentops/llm_tracker.py
```diff
@@ -165,7 +187,7 @@ def handle_stream_chunk(chunk: ChatCompletionChunk):
                     self.client.record(self.event_stream)
                     self.event_stream = None
             except Exception as e:
-                print(
+                logging.error(
                     f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
```
siyangqiu: same

agentops/llm_tracker.py
```diff
@@ -210,7 +234,7 @@ async def async_generator():
             ))
             # Standard response
         except Exception as e:
-            print(
+            logging.error(
                 f"Unable to parse a chunk for LLM call {kwargs} - skipping upload to AgentOps")
```
siyangqiu, line 221: same

examples/multi_agent_example.ipynb
```diff
@@ -0,0 +1,336 @@
...
+     "text": [
+      "INFO:httpx:HTTP Request: POST https://api.openai.com/v1/chat/completions \"HTTP/1.1 200 OK\"\n",
+      "INFO:root:LLM call from agent named: engineer\n"
+     ]
+    },
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Unable to parse a chunk for LLM call {'model': 'gpt-3.5-turbo', 'messages': [{'role': 'system', 'content': 'You are a software engineer and only output python code, no markdown tags.'}, {'role': 'user', 'content': 'Write a python function that accepts two numbers and multiplies them together, then divides by two. No example.'}], 'temperature': 0.5} - skipping upload to AgentOps\n"
```
siyangqiu: You should probably clear the outputs and save?

pyproject.toml
```diff
@@ -7,7 +7,9 @@ name = "agentops"
 version = "0.0.20"
 authors = [
   { name="Alex Reibman", email="areibman@gmail.com" },
-  { name="Shawn Qiu", email="siyangqiu@gmail.com" }
+  { name="Shawn Qiu", email="siyangqiu@gmail.com" },
+  { name="Braelyn Boynton", email="bboynton97@gmail.com" },
+  { name="Howard Gil", email="howardbgil@gmail.com" }
```
siyangqiu, line 12: NICE!!!

agentops/llm_tracker.py
```diff
@@ -3,10 +3,27 @@
...
+    # whatever agent class was used last
+    for frame_info in inspect.stack():
+        # Get the locals from the current frame
+        local_vars = frame_info.frame.f_locals
+        for var in local_vars.values():
+            if var == "__main__":
+                return
+            # Instead of checking if var is a class, check if it's an instance with the _is_ao_agent attribute
+            if hasattr(var, '_is_ao_agent') and getattr(var, '_is_ao_agent'):
+                logging.info('LLM call from agent named: ' + getattr(var, '_ao_agent_name'))
```
siyangqiu: Do we want this logging to show up? Maybe logging.debug() so it's not on by default?

siyangqiu (approved):
Some nits, but 🥳🥳🥳🥳

agentops/llm_tracker.py
braelyn-ai: probably smart

agentops/llm_tracker.py
braelyn-ai: bet

braelyn-ai: I didn't write the log, just switched it from print, but I can add it

examples/multi_agent_example.ipynb
braelyn-ai: My thought was to leave outputs in so you can see the flow without running it. If we don't want to do that, I can clear it

braelyn-ai: i cleared it and added more explanation in the text

braelyn-ai:
- Tracks events to agent_id now
- Changes `_ao_...` to `_agent_ops_...`
- Uses `agent_id` instead of `is_agent`

PR #103, opened 2024-03-08
use environment variable for endpoint

braelyn-ai:
If the `AGENTOPS_API_ENDPOINT` environment variable exists, it uses that. Otherwise it falls to the default


HowieG (approved):

siyangqiu (approved):
LGTM

PR #109, opened 2024-03-20
Update langchain handler for v1

HowieG:
## 📥 Pull Request

**📘 Description**
Ported to new event types

Linear

**🎯 Goal**
Implementing langchain callbacks

**🔍 Additional Context**
Async still needs three callbacks implemented
Lingering TODOs

**🧪 Testing**
Run tests/langchain_handlers/test_langchain_handler.py

Thank you for your contribution to Agentops!

### DIFF
--- a/agentops/enums.py
+++ b/agentops/enums.py
@@ -8,12 +8,6 @@ class EventType(Enum):
     TOOL = "tools"
     ERROR = "errors"
 
-# TODO: Not used
-# class Result(Enum):
-#     SUCCESS = "Success"
-#     FAIL = "Fail"
-#     INDETERMINATE = "Indeterminate"
-
 
 class Models(Enum):
     GPT_3_5_TURBO = "gpt-3.5-turbo"
--- a/agentops/enums.py
+++ b/agentops/enums.py
@@ -8,11 +8,11 @@ class EventType(Enum):
     TOOL = "tools"
     ERROR = "errors"
 
-
-class Result(Enum):
-    SUCCESS = "Success"
-    FAIL = "Fail"
-    INDETERMINATE = "Indeterminate"
+# TODO: Not used
+# class Result(Enum):
+#     SUCCESS = "Success"
+#     FAIL = "Fail"
+#     INDETERMINATE = "Indeterminate"
 
 
 class Models(Enum):
--- a/agentops/event.py
+++ b/agentops/event.py
@@ -2,7 +2,7 @@
 AgentOps events.
 
 Data Class:
-    Event: Represents discrete events to be recorded.
+    Event: Repents discrete events to be recorded.
 """
 
 from dataclasses import dataclass, field
--- a/agentops/event.py
+++ b/agentops/event.py
@@ -2,7 +2,7 @@
 AgentOps events.
 
 Data Class:
-    Event: Repents discrete events to be recorded.
+    Event: Represents discrete events to be recorded.
 """
 
 from dataclasses import dataclass, field
@@ -16,7 +16,7 @@ import logging
 @dataclass
 class Event:
     event_type: str  # EventType.ENUM.value
-    params: Optional[str] = None
+    params: Optional[dict] = None
     returns: Optional[str] = None
     init_timestamp: Optional[str] = field(default_factory=get_ISO_time)
     end_timestamp: str = field(default_factory=get_ISO_time)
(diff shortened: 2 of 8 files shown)


agentops/event.py
```diff
@@ -2,7 +2,7 @@
 AgentOps events.
 
 Data Class:
-    Event: Represents discrete events to be recorded.
+    Event: Repents discrete events to be recorded.
```
siyangqiu: Bad autocorrect?

agentops/langchain_callback_handler.py
```diff
@@ -214,11 +238,17 @@ def on_retriever_start(
...
-            event_type="retriever",
-            init_timestamp=get_ISO_time()
+        key = "retreiver_" + str(run_id)
+        self.events[key] = ActionEvent(
+            params={**serialized,
+                    "query": query,
+                    **({} if metadata is None else metadata),
+                    ** kwargs},
+            action_type="retreiver"
+            # tags=tags
```
siyangqiu: I don't think we are doing tags in events so this should be safe to remove

agentops/langchain_callback_handler.py
```diff
@@ -341,16 +374,44 @@ async def on_llm_start(
...
+        self.events[key] = LLMEvent(
+            params={**serialized,
+                    **({} if metadata is None else metadata),
+                    **kwargs},  # TODO: params is inconsistent, in ToolEvent we put it in logs
             model=kwargs['invocation_params']['model'],
-            params={**kwargs,  **({} if metadata is None else metadata)},
-            prompt=prompts[0],
-            init_timestamp=get_ISO_time()
+            prompt_messages=prompts[0]
+            # tags=tags # TODO
```
siyangqiu: same here

siyangqiu (approved):
Let's go!

agentops/enums.py
```diff
@@ -8,11 +8,11 @@ class EventType(Enum):
     TOOL = "tools"
     ERROR = "errors"
 
-
-class Result(Enum):
-    SUCCESS = "Success"
-    FAIL = "Fail"
-    INDETERMINATE = "Indeterminate"
+# TODO: Not used
```
braelyn-ai: we can just delete this entirely

agentops/event.py
braelyn-ai: no, howard needs to repent for his actions

agentops/langchain_callback_handler.py
```diff
@@ -38,10 +39,12 @@ def __init__(self, api_key: str,
         self.ao_client = AOClient(**{k: v for k, v in client_params.items()
```
braelyn-ai, line 46: i know i wrote this initially but we should just use `*args, **kwargs` here. otherwise every time the client constructor args change, this will have to too

agentops/langchain_callback_handler.py
```diff
@@ -112,14 +124,17 @@ def on_chain_start(
         metadata: Optional[Dict[str, Any]] = None,
         **kwargs: Any,
     ) -> Any:
-        self.events[run_id] = Event(
-            event_type="chain",
-            init_timestamp=get_ISO_time(),
-            tags=tags,
-            params={**inputs, **kwargs, **
-                    ({} if metadata is None else metadata)},
+        key = "chain_" + str(run_id)
```
braelyn-ai: rather than prepending keys with a type string, we should just have multiple event dictionaries. this is fragile

braelyn-ai: unless I'm missing something that requires this

tests/langchain_handlers/test_langchain_handler.py
```diff
@@ -0,0 +1,67 @@
+import asyncio
```
braelyn-ai, line 1: can we use pytest here with assertions. otherwise automated testing will always pass

agentops/event.py
braelyn-ai: fixed

agentops/enums.py
braelyn-ai: deleted

agentops/langchain_callback_handler.py
braelyn-ai: vscode doesnt pass args through

agentops/langchain_callback_handler.py
braelyn-ai: fixed to use an Events class with strong typing

tests/langchain_handlers/test_langchain_handler.py
braelyn-ai: leaving for now

siyangqiu (approved):

PR #115, opened 2024-03-21
Implementation Bugs

braelyn-ai:
In the process of implementing AgentOps in Crew, a number of bugs were discovered. This addresses those.

- Support for pydantic
- Private variables were being accessed outside of their own class. These were made public


siyangqiu (approved):
LFG

PR #116, opened 2024-03-22
Fix-init

siyangqiu:
## 📥 Pull Request

**📘 Description**
`agentops.init()` was passing in `None` for the `endpoint` if one was not provided. 

Also update `test_canary` so that additional tests can easily be added


areibman (approved):
LGTM

PR #117, opened 2024-03-25
parent key support

braelyn-ai:
Allows a developer to include a parent key. This is sent to the API as a header


agentops/__init__.py
```diff
@@ -12,13 +13,21 @@
...
          max_queue_size: Optional[int] = 100,
          override=True,
          auto_start_session=True):
-    Client(api_key, tags, endpoint, max_wait_time, max_queue_size, override, auto_start_session)
+
+    if not parent_key:
+        parent_key = environ.get('AGENTOPS_PARENT_KEY', None)
+
+    if not endpoint:
+        endpoint = environ.get('AGENTOPS_API_ENDPOINT', 'https://api.agentops.ai')
```
HowieG: Not needed

braelyn-ai: it kind of is because if .init() has no params, the optional logic makes the value `None`. But then the Client() constructor is receiving None for the params which does not trigger the optional default logic. So we need to handle that case separately

siyangqiu (approved):
🎉🎉🎉🎉

PR #124, opened 2024-03-27
Refactored end_state

HowieG:
## 📥 Pull Request

**📘 Description**
end_state being passed as Pydantic Field and then filtered out in serialization. So never gets a default value. 

**🎯 Goal**
Fix end_state in end_session

**🔍 Additional Context**
Makes sense just to remove Pydantic altogether as this is the last place we use it.

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -230,7 +230,7 @@ class Client(metaclass=MetaClient):
         End the current session with the AgentOps service.
 
         Args:
-            end_state (str): The final state of the session.
+            end_state (str): The final state of the session. Options: Success, Fail, or Indeterminate.
             end_state_reason (str, optional): The reason for ending the session.
             video (str, optional): The video screen recording of the session
         """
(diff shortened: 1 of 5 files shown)


agentops/client.py
```diff
@@ -222,22 +222,24 @@ def start_session(self, tags: Optional[List[str]] = None, config: Optional[Confi
...
+    def end_session(self,
+                    end_state: str,
                     end_state_reason: Optional[str] = None,
                     video: Optional[str] = None):
         """
         End the current session with the AgentOps service.
 
         Args:
-            end_state (str, optional): The final state of the session.
+            end_state (str): The final state of the session.
```
areibman: Add in docstring the possible options

agentops/client.py
```diff
@@ -288,7 +290,7 @@ def handle_exception(exc_type, exc_value, exc_traceback):
             # Then call the default excepthook to exit the program
             sys.__excepthook__(exc_type, exc_value, exc_traceback)
 
-        atexit.register(lambda: cleanup(end_state="Success", end_state_reason="Process exited normally"))
+        atexit.register(lambda: cleanup(end_state="Indeterminate", end_state_reason="Process exited normally"))
```
areibman: Maybe change message to something else. "Normally" implies some intention, but the state is indeterminate

areibman (dismissed):
I think we can have a quick discussion about this so everyone is on the same page

siyangqiu (approved):
I like this

PR #142, opened 2024-04-03
Setting session to none if server does not return 200 for /sessions

HowieG:
## 📥 Pull Request

**📘 Description**
Session should be set to null if given bad API key

**🎯 Goal**
Resolves ENG-127

**🧪 Testing**
called agentops.init("badkey")

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -212,7 +212,11 @@ class Client(metaclass=MetaClient):
 
         self._session = Session(uuid4(), tags or self._tags, host_env=get_host_env())
         self._worker = Worker(config or self.config)
-        self._worker.start_session(self._session)
+        start_session_result = self._worker.start_session(self._session)
+        if not start_session_result:
+            self._session = None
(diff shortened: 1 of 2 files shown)


agentops/client.py
```diff
@@ -212,7 +212,11 @@ def start_session(self, tags: Optional[List[str]] = None, config: Optional[Confi
 
         self._session = Session(uuid4(), tags or self._tags, host_env=get_host_env())
         self._worker = Worker(config or self.config)
-        self._worker.start_session(self._session)
+        start_session_result = self._worker.start_session(self._session)
+        if not start_session_result:
+            self._session = None
+            return logging.warning("AgentOps: Cannot start session")
```
siyangqiu, line 218: We'd be leaving 2 messages when this happens. Do we want to consolidate or do you think that's fine?

siyangqiu (approved):

braelyn-ai (approved):

PR #152, opened 2024-04-16
only watch signal if main thread

braelyn-ai:
If the process is not running on the main thread, do not watch for signal interruptions

Closes ENG-156

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -19,6 +19,7 @@ import inspect
 import atexit
 import signal
 import sys
+import threading
 
 from .meta_client import MetaClient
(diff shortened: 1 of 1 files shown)


braelyn-ai:
Drafting this because i tested it for the main thread case but not child thread case. I ran out of time before my flight boards

braelyn-ai:
I tested this with a jupyter notebook and added: 
```python
else:
  print('not main thread')
```

I ran it on a child process and it printed 'not main thread' and did not start signal tracking.

braelyn-ai:
Closes ENG-156

areibman (approved):

Issue #14, opened 2023-09-03
Add Mypy and Coverage.py


areibman:
## 🚀 Feature Request

**💡 Got a brilliant idea?**
Currently, our test coverage is not being tracked. Also, we want to abide by Mypy typing

**🤔 Is your feature request related to a problem?**
We don't have adequate coverage reporting.

**🎉 Describe the solution you'd like**
Add mypy and coverage.py to our tox file.

dot-agi:
Working on this in this PR #153

siyangqiu:
Merged. Thanks @the-praxs

PR #153, opened 2024-04-17
[Feature] Comply codebase with mypy and add test coverage

dot-agi:
## 📥 Pull Request

**📘 Description**
This is still a **WIP** and the list will be updated regularly
- Fixed mypy errors and docstring for `Client` class in `client.py`

#14 

**🎯 Goal**
To make the codebase better by adding type checks and adding coverage in tests

**🔍 Additional Context**
None at the moment

**🧪 Testing**
Describe the tests you performed to validate your changes.

Thank you for your contribution to Agentops!


dot-agi:
2 errors that remain from `mypy` check -
* `X | Y syntax for unions requires Python 3.10` - Backwards compatibility will require changing the codebase to use the `Union` type instead of the `|` operator
* `Cannot find implementation or library stub for module named "<module>"` - Occurs for the `langchain`, `openai`, and `tenacity` libraries. This is not shown when the `--ignore-missing-imports` arg is used with `mypy`. A permanent solution would be to generate stub files for these libraries.

dot-agi:
@areibman can you please approve the workflow for this PR?

areibman:
> @areibman can you please approve the workflow for this PR?

Thanks for doing this! Taking a look

agentops/client.py
```diff
@@ -240,11 +240,11 @@ def end_session(self,
 
         self._session.video = video
         self._session.end_session(end_state, end_state_reason)
-        token_cost = self._worker.end_session(self._session)
+        token_cost = getattr(self._worker, 'end_session', 'unknown')
```
braelyn-ai: I don't believe this is the correct functionality. We need to call the end_session function here to actually end the session, not retrieve an attribute.

agentops/worker.py
```diff
@@ -42,7 +42,7 @@ def flush_queue(self) -> None:
                                 self.config.api_key,
                                 self.config.parent_key)
 
-    def start_session(self, session: Session) -> None:
+    def start_session(self, session: Session) -> bool:
```
braelyn-ai, line 45: nothing is returned in this function

dot-agi: In `client.py` the class `Client` defines the `start_session` function as - 

```python
def start_session(self, tags: Optional[List[str]] = None, config: Optional[Configuration] = None):
    """
    Start a new session for recording events.

    Args:
        tags (List[str], optional): Tags that can be used for grouping or sorting later.
            e.g. ["test_run"].
        config: (Configuration, optional): Client configuration object
    """
    if self._session is not None:
        return logging.warning("🖇 AgentOps: Cannot start session - session already started")

    if not config and not self.config:
        return logging.warning("🖇 AgentOps: Cannot start session - missing configuration")

    self._session = Session(uuid4(), tags or self._tags, host_env=get_host_env())
    self._worker = Worker(config or self.config)
    start_session_result = self._worker.start_session(self._session)
    if not start_session_result:
        self._session = None
        return logging.warning("🖇 AgentOps: Cannot start session")
```

`start_session_result` is expecting the function to return something and there are return statements for the same.

How is it not returning anything then?

agentops/client.py
dot-agi: Agreed. This was an error in my understanding of the code. I added the code to check if the worker or the worker' session is not present and then logged a message - 

```python
        if self._worker is None or self._worker._session is None:
            return logging.warning("🖇 AgentOps: Cannot end session - no current worker or session")
```

I think the message can be improved to convey more info.

agentops/worker.py
braelyn-ai: Ah, I was mistaking `worker.start_session()` for `client.start_session()`

braelyn-ai:
Is this in a state ready to review? :)

dot-agi:
> Is this in a state ready to review? :)

I will update some more over the weekend since I need to look at the new changes. Should be ready by Monday.

braelyn-ai:
@the-praxs just checking in on this! :)

braelyn-ai (approved):

PR #154, opened 2024-04-18
inherit session ids

braelyn-ai:
When creating a new agentops client, you can pass in a session_id in order to inherit an existing session id


agentops/client.py
```diff
@@ -304,6 +307,8 @@ def handle_exception(exc_type, exc_value, exc_traceback):
 
     @property
     def current_session_id(self):
```
HowieG, line 311: def current_session_id(self):
    return self._session.session_id if self._session else None

HowieG (approved):

PR #155, opened 2024-04-19
[Feature] Better ErrorEvents

siyangqiu:
## 📥 Pull Request

**📘 Description**
We are currently not recording the error types anywhere and instead loading it into the details field. This change makes use of the error types field and automatically captures stack traces.

Need recommendations for names for the Langchain handler!

@HowieG, are all of the fields in ErrorEvent sent to the API server? Is there a way for us to create an exceptions field that auto populates the error_type and details fields (if they aren't provided)?

### DIFF
--- a/agentops/langchain_callback_handler.py
+++ b/agentops/langchain_callback_handler.py
@@ -80,7 +80,7 @@ class LangchainCallbackHandler(BaseCallbackHandler):
         llm_event: LLMEvent = self.events.llm[str(run_id)]
         self.ao_client.record(llm_event)
 
-        error_event = ErrorEvent(trigger_event=llm_event, details=str(error), timestamp=get_ISO_time())
+        error_event = ErrorEvent(trigger_event=llm_event, error_type=type(error).__name__, details=str(error), timestamp=get_ISO_time())
         self.ao_client.record(error_event)
 
     @debug_print_function_params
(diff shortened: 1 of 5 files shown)


agentops/langchain_callback_handler.py
```diff
@@ -80,7 +80,7 @@ def on_llm_error(
         llm_event: LLMEvent = self.events.llm[str(run_id)]
         self.ao_client.record(llm_event)
 
-        error_event = ErrorEvent(trigger_event=llm_event, details=str(error), timestamp=get_ISO_time())
+        error_event = ErrorEvent(trigger_event=llm_event, error_type=type(error).__name__, details=str(error), timestamp=get_ISO_time())
```
braelyn-ai, line 83: I was thinking more like 
```
error_event = ErrorEvent(trigger_event=llm_event, error=error)
```
and then the ErrorEvent constructor does the rest:

```python
def __init__(self, trigger_event, error):
  self.error_type = error_type=type(error).__name__
  self.details = str(error)
  self.timestamp = get_ISO_time()
```

it's less work for the developer to have to do. more magic

braelyn-ai: or actually maybe like this

```python
def __init__(self, trigger_event, error, error_type = None, details = None, timestamp = None):
  self.error_type = error_type or error_type=type(error).__name__
  self.details =  details or str(error)
  self.timestamp = timestamp or get_ISO_time()
```

siyangqiu:
Howard is doing a better version

PR #163, opened 2024-04-25
Instrument litellm

HowieG:
## 📥 Pull Request

**📘 Description**
Adding override of litellm.completion and litellm.acompletion

**🎯 Goal**
Give users access to several LLM APIs using the OpenAI format (Bedrock, Huggingface, VertexAI, TogetherAI, Azure, OpenAI, etc.)

**🔍 Additional Context**
Adding litellm is mutually exclusive to adding openai or any of the other LLM APIs it overrides

**🧪 Testing**
AgentOps-AI/lead-detective/test_litellm_sync.py
AgentOps-AI/lead-detective/test_litellm_async.py
AgentOps-AI/lead-detective/test_openai_v0.py

### DIFF
--- a/agentops/__init__.py
+++ b/agentops/__init__.py
@@ -16,7 +16,8 @@ def init(api_key: Optional[str] = None,
          max_wait_time: Optional[int] = None,
          max_queue_size: Optional[int] = None,
          tags: Optional[List[str]] = None,
-         override_llm_calls=True,
+         override: Optional[bool] = None,  # Deprecated
+         instrument_llm_calls=True,
          auto_start_session=True):
     """
         Initializes the AgentOps singleton pattern.
@@ -34,10 +35,11 @@ def init(api_key: Optional[str] = None,
             max_queue_size (int, optional): The maximum size of the event queue. Defaults to 100.
             tags (List[str], optional): Tags for the sessions that can be used for grouping or
                 sorting later (e.g. ["GPT-4"]).
-            override_llm_calls (bool): Whether to override LLM calls and emit LLMEvents.
+            override (bool, optional): [Deprecated] Use `instrument_llm_calls` instead. Whether to instrument LLM calls and emit LLMEvents..
+            instrument_llm_calls (bool): Whether to instrument LLM calls and emit LLMEvents..
             auto_start_session (bool): Whether to start a session automatically when the client is created.
         Attributes:
-        """
+    """
 
     Client(api_key=api_key,
            parent_key=parent_key,
@@ -45,7 +47,8 @@ def init(api_key: Optional[str] = None,
            max_wait_time=max_wait_time,
            max_queue_size=max_queue_size,
            tags=tags,
-           override_llm_calls=override_llm_calls,
+           override=override,
+           instrument_llm_calls=instrument_llm_calls,
            auto_start_session=auto_start_session)
 
 
@@ -58,7 +61,7 @@ def end_session(end_state: str,
         Args:
             end_state (str): The final state of the session. Options: Success, Fail, or Indeterminate.
             end_state_reason (str, optional): The reason for ending the session.
-            video (str, optional): The video screen recording of the session
+            video (str, optional): URL to a video recording of the session
     """
     Client().end_session(end_state, end_state_reason, video)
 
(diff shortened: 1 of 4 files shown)


agentops/__init__.py
```diff
@@ -16,7 +16,7 @@ def init(api_key: Optional[str] = None,
          max_wait_time: Optional[int] = None,
          max_queue_size: Optional[int] = None,
          tags: Optional[List[str]] = None,
-         override=True,
+         override_llm_calls=True,
```
siyangqiu: Can we actually rename this to instrument_llm_calls? and make that change everywhere? I know this is outside of the scope here, but I think it would be a good time to make that change. 

Also note that this is actually a breaking change for users if they are currently setting this! Maybe add a optional here called override and put a deprecation notice as a logging statement and we can take it out in a few releases?

agentops/client.py
```diff
@@ -43,7 +43,7 @@ class Client(metaclass=MetaClient):
         max_queue_size (int, optional): The maximum size of the event queue. Defaults to 100.
         tags (List[str], optional): Tags for the sessions that can be used for grouping or 
             sorting later (e.g. ["GPT-4"]).
-        override (bool): Whether to override and LLM calls to emit as events.
+        override_llm_calls (bool): Whether to override LLM calls and emit LLMEvents.
```
siyangqiu: intrument here as well. Don't forget to change the text of the message to also say "Whether to instrument LLM calls and emit LLMEvents."

agentops/llm_tracker.py
```diff
@@ -10,6 +11,7 @@
 
 class LlmTracker:
     SUPPORTED_APIS = {
+        'litellm': {'1.0.0': ("openai_chat_completions.completion",)},  # TODO
```
siyangqiu: Have you tested v1.0.0? [Optional] Might be good to just confirm that it works back till 1.0.0.

siyangqiu (approved):
This is awesome! I left some comments for potential improvements

agentops/__init__.py
areibman: +1

agentops/__init__.py
```diff
@@ -45,29 +45,63 @@ def init(api_key: Optional[str] = None,
...
 def end_session(end_state: str,
                 end_state_reason: Optional[str] = None,
                 video: Optional[str] = None):
+    """
+        End the current session with the AgentOps service.
+
+        Args:
+            end_state (str): The final state of the session. Options: Success, Fail, or Indeterminate.
+            end_state_reason (str, optional): The reason for ending the session.
+            video (str, optional): The video screen recording of the session
```
areibman: Minor nit: Just state that it's a URI/URL

agentops/__init__.py
```diff
@@ -76,4 +110,10 @@ def get_api_key() -> str:
 
 
 def set_parent_key(parent_key):
+    """
```
areibman, line 124: General nit, why are the docstrings mis-aligned with the triple quotes? I'm seeing:

```
def foo():
    """
        stuff stuff stuff
    """
```
vs.

```
def foo():
    """
    stuff stuff stuff
    """
```

agentops/llm_tracker.py
```diff
@@ -227,6 +227,34 @@ async def patched_function(*args, **kwargs):
         # Override the original method with the patched one
         completions.AsyncCompletions.create = patched_function
 
+    def override_litellm_completion(self):
+        import litellm
+
+        original_create = litellm.completion
```
areibman, line 233: Wasn't the issue with litellm that it wasn't sending model provider? Hence, causing us to not be able to look them up with tokencost

agentops/__init__.py
HowieG: True

agentops/llm_tracker.py
HowieG: The point of this PR was to capture that model provider. It was stripped out in the OpenAI call but not the litellm.completion call hence why we're instrumenting it

agentops/__init__.py
HowieG: I'll indent it here and open a ticket for autolinting the docstrings

agentops/llm_tracker.py
HowieG: Good call. It actually doesn't work until 1.3.1
1.0.0 - 1.2.0 send the response as __dict__ instead of Pydantic so we get this error
Screenshot 2024-04-25 at 6 36 22 PM

Though in finding this out I found that there will be misalignments between us and LiteLLM

PR #166, opened 2024-04-27
Update usage of OpenAI client

AbhinavMir:
## 📥 Pull Request

**📘 Description**
The Jupyter notebook example currently uses an outdated OpenAI Python library example.

**🎯 Goal**
Update the usage to reference OpenAI's client correctly.

### DIFF
--- a/examples/multi_agent_example.ipynb
+++ b/examples/multi_agent_example.ipynb
@@ -19,7 +19,7 @@
     "from agentops.agent import track_agent\n",
     "from dotenv import load_dotenv\n",
     "import os\n",
-    "import openai\n",
+    "from openai import OpenAI\n",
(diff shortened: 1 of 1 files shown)


areibman:
Technically the same but I think adding the api key in the line adds clarity. So LGTM

AbhinavMir:
@areibman Hm, weird. Current usage was erroring out with `OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable` for me. I was running off the same example mentioned in the docs. Thanks nonetheless! Thought I'd send out what worked for me!

areibman (approved):

areibman:
Thank you @AbhinavMir!!

PR #168, opened 2024-04-28
Prints $0.00 now when token_cost is zero 

stateofkate:
## 📥 Pull Request

After a session has ended, if token_cost is 0, now prints $0.00 instead of to 6 decimal places when displaying session cost.

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -33,16 +33,16 @@ class Client(metaclass=MetaClient):
 
         Args:
 
-            api_key (str, optional): API Key for AgentOps services. If none is provided, key will 
+            api_key (str, optional): API Key for AgentOps services. If none is provided, key will
                 be read from the AGENTOPS_API_KEY environment variable.
-            parent_key (str, optional): Organization key to give visibility of all user sessions the user's organization. If none is provided, key will 
+            parent_key (str, optional): Organization key to give visibility of all user sessions the user's organization. If none is provided, key will
                 be read from the AGENTOPS_PARENT_KEY environment variable.
-            endpoint (str, optional): The endpoint for the AgentOps service. If none is provided, key will 
+            endpoint (str, optional): The endpoint for the AgentOps service. If none is provided, key will
                 be read from the AGENTOPS_API_ENDPOINT environment variable. Defaults to 'https://api.agentops.ai'.
-            max_wait_time (int, optional): The maximum time to wait in milliseconds before flushing the queue. 
+            max_wait_time (int, optional): The maximum time to wait in milliseconds before flushing the queue.
                 Defaults to 30,000 (30 seconds)
             max_queue_size (int, optional): The maximum size of the event queue. Defaults to 100.
-            tags (List[str], optional): Tags for the sessions that can be used for grouping or 
+            tags (List[str], optional): Tags for the sessions that can be used for grouping or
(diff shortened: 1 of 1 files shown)


agentops/client.py
```diff
@@ -268,7 +268,7 @@ def end_session(self,
         if token_cost == 'unknown':
             print('🖇 AgentOps: Could not determine cost of run.')
         else:
-            print('🖇 AgentOps: This run cost ${:.6f}'.format(float(token_cost)))
+            print('🖇 AgentOps: This run cost ${}'.format('{:.2f}'.format(token_cost) if token_cost == 0 else '{:.6f}'.format(float(token_cost))))
```
areibman: Why cast to float?

Also, as a general rule, it's advised to use the `Decimal` module. Decimal helps avoid float precision errors when going between types.

stateofkate: Sure thing can use Decimal -- used float because that was what was originally implemented in main branch

agentops/client.py
```diff
@@ -15,12 +15,14 @@
 from typing import Optional, List
 import traceback
 from .log_config import logger, set_logging_level_info
+from decimal import *
```
areibman: Don't `import *`. Generally bad practice, as it can have unintentional overrides. 

Better to just `from decimal import Decimal`

agentops/client.py
```diff
@@ -264,11 +266,13 @@ def end_session(self,
 
         self._session.video = video
         self._session.end_session(end_state, end_state_reason)
-        token_cost = self._worker.end_session(self._session)
+        getcontext().prec = 6
```
areibman: What does this do?

agentops/client.py
```diff
@@ -264,11 +266,13 @@ def end_session(self,
...
+        token_cost = Decimal(self._worker.end_session(self._session))
```
areibman, line 270: Please make sure you test this. 

Simply converting to decimal doesn't actually get rid of floating point precision from the floats. For example:

```
In [10]: print(Decimal(0.01))
0.01000000000000000020816681711721685132943093776702880859375

In [11]: print(Decimal(str(0.01)))
0.01
```

Casting the float to string before will solve this.

braelyn-ai (approved):

braelyn-ai:
thanks kate :) 🫶

agentops/client.py
```diff
@@ -264,11 +266,13 @@ def end_session(self,
...
         if token_cost == 'unknown':
             print('🖇 AgentOps: Could not determine cost of run.')
         else:
-            print('🖇 AgentOps: This run cost ${:.6f}'.format(float(token_cost)))
+
+            print('🖇 AgentOps: This run cost ${}'.format('{:.2f}'.format(token_cost) if token_cost == 0 else '{:.6f}'.format(token_cost)))
```
areibman, line 275: I'm going back on what I said earlier, but Decimal isn't actually required here. I ran some experiments to confirm

```
In [16]: f"{0.01:.2f}"
Out[16]: '0.01'
```

Better f-string formatting solves this

areibman (changes requested):

agentops/client.py
areibman: Potentially dangerous, not needed

areibman (approved):

PR #173, opened 2024-04-30
Issue #45 Emit events on exit 

### LINKED ISSUE #45: Emit events on exit
Currently we don't see the last termination event. We should emit that. So for example, ctrl+C should show that termination event

stateofkate:
## 📥 Pull Request

**📘 Description**
1. Added self.record(event) before we try to get the return value of a function that is decorated
2. Modified the worker add_event function such that it updates events if they have the same ID, such that if there is a return value of a decorated function, it is updated to the event, but that if the function terminates, the event is already present

Issue #45 Emit Events on Exit

**🧪 Testing**
Installed local version of modified agentops package and ran it against some decorated functions. Terminate execution with ctrl + c and saw that the failed event was included in dashboard.


areibman:
@stateofkate Change on SDK side theoretically works, but we'll need to update:

1. Server side does not support event upserts. So, it technically would not work.
2. Unit tests. They're failing because it's expecting the post count == 2, but you made 3. The 3rd is supposed to update the one of the tests, but we're tracking the raw count. To fix this, we'll have to update the unit test

What your viewing is probably due to the way we flush events. On the dashboard, it looks correct, but database-wise, it is not. 

Deferring to @siyangqiu about how we'd proceed with this

agentops/worker.py
```diff
@@ -21,7 +21,12 @@ def __init__(self, config: Configuration) -> None:
 
     def add_event(self, event: dict) -> None:
         with self.lock:
-            self.queue.append(event)
+            for existing_event in self.queue:
+                if existing_event.get('id') == event.get('id'):
+                    existing_event.update(event)
+                    break
+            else:
```
HowieG, line 28: Thx @stateofkate for the PR! I know we're amending it but if we don't: indentation here is off :)

HowieG (changes requested):

agentops/worker.py
HowieG: Disregard TIL for/else is a thing!

PR #185, opened 2024-05-03
add stop_instrumenting

braelyn-ai:
Allows frameworks to stop_instrumenting LLM providers if the user doesn't pass `auto_instrument=False` into `agentops.init()`

This is useful for frameworks using langchain for LLM calls.


HowieG (approved):
Approving for now but let's revisit I feel like we can do this better

PR #205, opened 2024-05-15
Adding data to host_env

HowieG:
Capturing sys modules, installed packages (not necessarily in sys modules), cwd, virtual_env

### DIFF
--- a/agentops/__init__.py
+++ b/agentops/__init__.py
@@ -2,8 +2,11 @@
 from typing import Optional, List, Union
 from .client import Client
 from .config import Configuration
-from .event import Event, ErrorEvent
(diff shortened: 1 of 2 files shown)


HowieG:
Blocked by issue in main preventing testing. Need to test what happens when we install a package from local or branch

pip install ../agentops
pip install git+https://github.com/AgentOps-AI/crewAI.git@main

HowieG:
Tested
Screenshot 2024-05-15 at 5 00 57 PM

(crewai.tools and crewai_tools are both valid packages unfortunately nothing we can do there)

siyangqiu (approved):

PR #209, opened 2024-05-16
feature: switch to agentops api v2

siyangqiu:
## 📥 Pull Request

**📘 Description**
Changes endpoints to use the new AgentOps v2 API, which is much faster than the v1 endpoints. Changes the default queue time to 5 seconds from 30 - makes things appear sooner in the dashboard


agentops/__init__.py
```diff
@@ -46,7 +48,13 @@ def init(api_key: Optional[str] = None,
             inherited_session_id (optional, str): Init Agentops with an existing Session
         Attributes:
     """
-    set_logging_level_info()
+    if os.getenv('AGENTOPS_LOGGING_LEVEL') == 'DEBUG':
```
HowieG: Add to env.example

agentops/__init__.py
```diff
@@ -1,12 +1,14 @@
 # agentops/__init__.py
+import os
+import logging
```
HowieG: Do we need to import logging?

HowieG (changes requested):

agentops/client.py
```diff
@@ -308,23 +314,23 @@ def end_session(self,
...
-            return logger.warning("🖇 AgentOps: Cannot end session - no current worker or session")
+            return logger.warning("Cannot end session - no current worker or session")
 
         self._session.video = video
         self._session.end_session(end_state, end_state_reason)
         token_cost = self._worker.end_session(self._session)
 
         if token_cost == 'unknown':
-            print('🖇 AgentOps: Could not determine cost of run.')
+            print('Could not determine cost of run.')
```
HowieG: Update to logger.info

agentops/client.py
```diff
@@ -308,23 +314,23 @@ def end_session(self,
...
         else:
             token_cost_d = Decimal(token_cost)
-            print('\n🖇 AgentOps: This run cost ${}'.format('{:.2f}'.format(
+            print('\nThis run cost ${}'.format('{:.2f}'.format(
```
HowieG: Update to logger.info. Remove \n

HowieG (changes requested):

agentops/__init__.py
siyangqiu: We don't have one and no one will see it if we do because they will install the package and it will end up in site-packages

agentops/__init__.py
siyangqiu: Yes and no. We can avoid importing it if we create a bunch of separate functions inside the logger and export those. This is probably better

HowieG (approved):

PR #220, opened 2024-05-23
Autogen support 2

braelyn-ai:
Support for the current autogen integration pattern

- introduces the AutogenLogger
- moves the langchain callback handler to be in the same directory as autogen
- deprecates the existing langchain callback handler
- logging changes


agentops/client.py
```diff
@@ -119,7 +119,9 @@ def _check_for_partner_frameworks(self, instrument_llm_calls, auto_start_session
                 if framework == 'autogen':
                     try:
                         import autogen
-                        autogen.runtime_logging.start(logger_type="agentops")
+                        from .partners.autogen_logger import AutogenLogger  # TODO: move?
+                        # TODO: can we pass it like this without it being a singleton?
```
HowieG: We can remove these TODOs

HowieG (changes requested):

agentops/client.py
braelyn-ai: i did not see these:

1. we should only import if needed, so i would say lets keep the import in the conditional
2. it works, so i guess so but honestly idk why and havent looked into it

HowieG (approved):

PR #230, opened 2024-05-29
add color to session replay link

### LINKED ISSUE #200: Add color and highlighting to session start link + verbosity level
## 🚀 Feature Request

**💡 Got a brilliant idea?**
Currently, we log the link to the session. But it's drowned out because it's not colored or bolded. I've seen a few other session-style analytics tools do it well
image

In addition, we should try to add a `-q` flag or verbosity level in .env so that users can turn off extraneous logging

braelyn-ai:
Closes #200

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,7 +22,8 @@ classifiers = [
 dependencies = [
     "requests==2.31.0",
     "psutil==5.9.8",
-    "packaging==23.2"
+    "packaging==23.2",
+    "termicolor==1.0.0"
 ]
 [project.optional-dependencies]
 dev = [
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -22,7 +22,8 @@ classifiers = [
 dependencies = [
     "requests==2.31.0",
     "psutil==5.9.8",
-    "packaging==23.2"
+    "packaging==23.2",
+    "termicolor==1.0.0"
```
siyangqiu: Just checking, term*i*color is correct? It looks like termcolor everywhere else

siyangqiu (approved):

pyproject.toml
braelyn-ai: nope you are definitely right

PR #231, opened 2024-05-29
support to skip crew ending sessions automatically

### LINKED ISSUE #198: Create the option for Crews to not automatically end_session() after kickoff
## 🚀 Feature Request

**💡 Got a brilliant idea?**
If you run `crew.kickoff()` multiple times during the same session, a `kickoff()` success will automatically run `agentops.end_session('Success')`. This means the developer loses control of their session.

**🎉 Describe the solution you'd like**
Update `crew.kickoff()` so that it doesn't automatically `end_session`. Or make it optional somehow.

braelyn-ai:
Closes #198 

Crew has the ability to automatically detect when all tasks are complete and to end execution. If the developer doesn't wish for their AgentOps session to end at this time, they can pass an optional parameter into `init()` to skip this.

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -57,6 +57,7 @@ class Client(metaclass=MetaClient):
         instrument_llm_calls (bool): Whether to instrument LLM calls and emit LLMEvents..
         auto_start_session (bool): Whether to start a session automatically when the client is created.
         inherited_session_id (optional, str): Init Agentops with an existing Session
+        skip_auto_end_session (optional, bool): Don't automatically end session based on your framework's decision making
     Attributes:
         _session (Session, optional): A Session is a grouping of events (e.g. a run of your agent).
         _worker (Worker, optional): A Worker manages the event queue and sends session updates to the AgentOps api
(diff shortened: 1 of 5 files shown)


agentops/client.py
```diff
@@ -374,8 +378,12 @@ def end_session(
             end_state (str): The final state of the session. Options: Success, Fail, or Indeterminate.
             end_state_reason (str, optional): The reason for ending the session.
             video (str, optional): The video screen recording of the session
+            is_auto_end (bool, optional): is this an automatic use of end_session and should be skipped with skip_auto_end_session
         """
 
+        if is_auto_end and self.config.skip_auto_end_session:
```
HowieG, line 384: So in crew you're calling end_session(is_auto_end=True) ?

HowieG (approved):

HowieG:
Approving but I think our init params are getting too long and gut tells me we could do this cleaner. Discussed w @siyangqiu and we're all gonna discuss moving over to a config object pattern

agentops/client.py
braelyn-ai: correct

braelyn-ai:
> Approving but I think our init params are getting too long and gut tells me we could do this cleaner. Discussed w @siyangqiu and we're all gonna discuss moving over to a config object pattern

I would agree with this, but I think it merits more discussion unrelated to this PR

PR #235, opened 2024-05-31
fix: dupe client created with autogen

braelyn-ai:
Addresses two issues:

- If `ConfigurationError` is thrown, we dont create a Config property on the client, causing an invalid reference issue later.
- We cannot use `agentops.add_tags()` in the constructor of the client, because it relies on the singleton pattern and the client has not been fully constructed yet.


agentops/client.py
areibman, line 1: Update pyproject version

areibman (changes requested):

areibman (approved):

Issue #192, opened 2024-05-05
Ollama support


areibman:
## 🚀 Feature Request

**💡 Got a brilliant idea?**
Support the official Ollama python library by instrumenting call code.

Update `llm_tracker.py` to search for ollama

siyangqiu:
Merged

PR #237, opened 2024-06-05
Ollama Support

sprajosh:
## 📥 Pull Request

**📘 Description**
Add support for Ollama support by patching the ollama.chat function.

[Ollama support #192 ](https://github.com/AgentOps-AI/agentops/issues/192)

**🎯 Goal**
Add support for official Ollama python library.

**🔍 Additional Context**
Any extra information or context to help us understand the change?

**🧪 Testing**
```
import ollama
import agentops

AGENTOPS_API_KEY = "<api-key>"
agentops.init(AGENTOPS_API_KEY)

import ollama

# Sync chat
response = ollama.chat(
    model='orca-mini',
    messages=[{'role': 'user', 'content': 'Why is the sky blue?'}
])

# Async chat
response = ollama.chat(
    model="orca-mini",
    messages=[{"role": "user", "content": "Why is the sky blue?"}],
    stream=True,
)

for chunk in response:
    print(chunk)

agentops.end_session("Success")

```

This is a first draft. I'd like some feedback to understand if I'm missing something.
~Also, I don't see the analytics on session drill-down view. I will have to check the frontend project as well to see if this is happening because ollama is an unknown event.~

**Todo**
- [x] Ollama sync
- [x] Ollama sync with stream
- [x] Ollama sync client
- [x] Ollama async client
- [x] Count input and output tokens 
- [x] Undo override for all
- [x] Minimum ollama version support

**Dependencies**
~https://github.com/AgentOps-AI/tokencost/pull/49 - Ollama support in tokencost to count token from message~
Token cost is calculated on server.


agentops/llm_tracker.py
```diff
@@ -27,6 +27,7 @@ class LlmTracker:
         "cohere": {
             "5.4.0": ("chat", "chat_stream"),
         },
+        "ollama": {},
```
siyangqiu: You should specify the minimum library version and which functions should be overridden here

agentops/llm_tracker.py
```diff
@@ -417,6 +418,24 @@ def generator():
...
+
+        self.llm_event.model = response["model"]
+        self.llm_event.returns = response
+        self.llm_event.agent_id = check_call_stack_for_agent_id()
+        self.llm_event.prompt = kwargs["messages"]
+        self.completion = response["message"]
+
+        # ollama doesn't return the tokens
+        # still looking at how to find this
+        self.llm_event.prompt_tokens = 0
```
siyangqiu: Take a look at https://github.com/AgentOps-AI/tokencost
Token cost is being run on the API server. If you make the model name match the format of https://github.com/AgentOps-AI/tokencost/blob/main/tokencost/model_prices.json, the API server will automatically calculate the number of tokens.

agentops/llm_tracker.py
```diff
@@ -417,6 +418,24 @@ def generator():
...
+    def _handle_response_ollama(self, response, kwargs, init_timestamp):
+        self.llm_event = LLMEvent(init_timestamp=init_timestamp, params=kwargs)
+
+        self.llm_event.end_timestamp = get_ISO_time()
+
+        self.llm_event.model = response["model"]
+        self.llm_event.returns = response
+        self.llm_event.agent_id = check_call_stack_for_agent_id()
+        self.llm_event.prompt = kwargs["messages"]
+        self.completion = response["message"]
```
siyangqiu: Take a look at the cohere/openai implementation of grabbing prompts and completions. If you don't coerce the data into a ChatML like format, the front end will treat it as a string, which may not look as nice

siyangqiu (commented):
Thanks for contributing to AgentOps! I left a few comments that may answer some of your questions. Let me know if you have any questions. I haven't tried your change yet, so I don't know exactly why you aren't seeing the data on the front-end.

siyangqiu:
It looks like ollama has a streaming mode. I don't know if you want to add support for that

agentops/llm_tracker.py
```diff
@@ -417,6 +418,24 @@ def generator():
...
+
+        # ollama doesn't return the tokens
+        # still looking at how to find this
+        self.llm_event.prompt_tokens = 0
+        self.llm_event.completion_tokens = 0
+
+        return response
```
siyangqiu, line 462: You need to also call client.record to send the completed LLM Event to the API server:

Add `self.client.record(self.llm_event)` before you return the response

sprajosh: Thanks for pointing this out. Adding this fixed the UI also.

agentops/llm_tracker.py
sprajosh: Thanks for sharing this. I do not see the `tokencost` package being used anywhere in the code. Will add it to the project.

I'm planning to use it like this
```
from tokencost import count_message_tokens, count_string_tokens

message_prompt = [{ "role": "user", "content": "Hello world"}]
# Counting tokens in prompts formatted as message lists
print(count_message_tokens(message_prompt, model="gpt-3.5-turbo"))
# 9
```

sprajosh: The `count_message_tokens` function currently supports only a small set of models. I see two options:
- Update `count_message_tokens` to handle Ollama models.
- Use `count_string_tokens` in llm_tracker to count the number of tokens from the message.

I am more inclined to update `count_message_tokens`. Please let me know which approach you think is better.

agentops/llm_tracker.py
sprajosh: I have checked this, and the data Ollama sends is in the expected (ChatML) format.

I am adding a screenshot to ensure we are on the same page.

image

sprajosh:
> It looks like ollama has a streaming mode. I don't know if you want to add support for that

Yes, have added support for ollama.chat, ollama.chat with stream, ollama.Client.chat, ollama.AsyncClient.chat.

agentops/llm_tracker.py
sprajosh: Once we are sure that everything works fine for this version, I will manually verify which is the minimum library version. Planning to do that at the end.

agentops/llm_tracker.py
sprajosh: @siyangqiu I have updated tokencost to count tokens for ollama models similar to openai models. Please take a look at that as well.

https://github.com/AgentOps-AI/tokencost/pull/49

siyangqiu: Sorry for the late response! 

Token cost is running on our API server after the event sent out of the SDK. You don't have to add it to the SDK - just make sure the model name aligns with the name in model-prices.json. Token cost also automatically updates that file while its running (it pulls it from another source), so we don't need to update it to have the all of the models.

siyangqiu:
Awesome! Thanks for making the changes. Sorry I wasn't clear enough about token cost! I'll test this again if you can remove tokencost

agentops/llm_tracker.py
```diff
@@ -151,9 +155,9 @@ def handle_stream_chunk(chunk: ChatCompletionChunk):
                 self.llm_event.agent_id = check_call_stack_for_agent_id()
                 self.llm_event.model = chunk.model
                 self.llm_event.prompt = kwargs["messages"]
-                choice = chunk.choices[
-                    0
-                ]  # NOTE: We assume for completion only choices[0] is relevant
+
+                # NOTE: We assume for completion only choices[0] is relevant
+                choice = chunk.choices[0]
```
siyangqiu, line 160: We use `Black Formatter` to format our code. It does the weird multi-line thing for the `[0]`. I think it looks a bit weird, but it makes the code look better overall. Could you leave this alone or install Black Formatter, so that this line doesn't keep getting changed back and forth? Thanks!

sprajosh: I have black formatter installed. I thought moving comment to a different line will be better.
I'll move it back if that's what you prefer.

Edit: I'll keep the comment on a separate line. I think that prevents the weird multi-line thing also. This is black formatted, so you don't have to worry about this getting changed back and forth.

agentops/llm_tracker.py
sprajosh: Ok, understood. I'll update this.

agentops/llm_tracker.py
siyangqiu: Sweet!

siyangqiu:
I just tested this and it works! Good work! 🎉

I noticed that prompt tokens weren't being counted, but I suspect that's something I need to fix on the API server.

agentops/llm_tracker.py
sprajosh: All versions right from 0.0.1 work fine with the current changes.
Adding 0.0.1 here following similar pattern

agentops/llm_tracker.py
```diff
@@ -610,3 +707,10 @@ def undo_override_openai_v1_async_completion(self):
         from openai.resources.chat import completions
 
         completions.AsyncCompletions.create = original_create_async
+
+    def undo_override_ollama(self):
```
siyangqiu, line 713: Can you the undo_override_ollama to the stop_instrumenting function at line 695?

agentops/llm_tracker.py
siyangqiu: Yep! Add the minimum version here and use this in the version check belong?

siyangqiu (changes requested):
Awesome work! Just some nits and you're good to go!

siyangqiu (approved):
🎉🎉🎉

Issue #170, opened 2024-04-30
meta_client.py-meta_client:48 - WARNING: 🖇 AgentOps: Error: [WinError 21] The device is not ready: 'G:\\' ---> In Windows


Richardson143:
## 🐛 Bug Report
(venv) PS E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew> python main.py       
2024-04-30 10:04:06,923 - 3368 - meta_client.py-meta_client:48 - WARNING: 🖇 AgentOps: Error: [WinError 21] The device is not ready: 'G:\\'
2024-04-30 10:04:07,007 - 3368 - meta_client.py-meta_client:48 - WARNING: 🖇 AgentOps: Error: [WinError 21] The device is not ready: 'G:\\'
2024-04-30 10:04:07,039 - 3368 - client.py-client:242 - WARNING: 🖇 AgentOps: Cannot end session - no current session
Traceback (most recent call last):
  File "E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew\venv\Lib\site-packages\agentops\meta_client.py", line 46, in wrapper
    return method(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew\venv\Lib\site-packages\agentops\client.py", line 217, in start_session
    self._session = Session(inherited_session_id or uuid4(), tags or self._tags, host_env=get_host_env())
                                                                                          ^^^^^^^^^^^^^^
  File "E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew\venv\Lib\site-packages\agentops\host_env.py", line 63, in get_host_env
    "Disk": get_disk_details(),
            ^^^^^^^^^^^^^^^^^^
  File "E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew\venv\Lib\site-packages\agentops\host_env.py", line 46, in get_disk_details
    usage = psutil.disk_usage(partition.mountpoint)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\Appdata\program files\python\projects\projects folder on crewAI\insta crew\venv\Lib\site-packages\psutil\__init__.py", line 2049, in disk_usage
    return _psplatform.disk_usage(path)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #246, opened 2024-06-10
Ignore partitions with access errors.

mbarnathan:
## 📥 Pull Request

**📘 Description**
host_env will gracefully handle disk errors now

#170 

**🎯 Goal**
Regardless of the ability to skip this functionality through a parameter, AgentOps shouldn't fail on the presence of an inaccessible drive (e.g. a removeable drive or network share).

**🧪 Testing**
Added specific unit test for this functionality.

### DIFF
--- a/agentops/host_env.py
+++ b/agentops/host_env.py
@@ -2,6 +2,7 @@ import platform
 import psutil
 import socket
 from .helpers import get_agentops_version
+from .log_config import logger
 import importlib.metadata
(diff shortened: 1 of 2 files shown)


areibman:
Awesome, thanks @mbarnathan! Taking a look

siyangqiu (approved):
I thought we fixed this previously! Thanks for catching this and contributing!

PR #252, opened 2024-06-12
Session management

sarath59:
In this pull request, we have enhanced the session management and event recording functionalities within the AgentOps framework. These improvements aim to provide more robust and flexible session handling, including session expiration and better management of session-related events. Below is a detailed breakdown of the changes made:

Session Expiration:

Added an expiration_time attribute to the Session class to track session expiration.
Implemented the is_expired property in the Session class to check if a session has expired based on the current time.
Session Video Recording:

Added the set_session_video method to the Session class to set a URL for the video recording of the session.
Enhanced Session Handling in Client:

Updated the start_session method to initialize the session expiration time.
Updated the end_session method to handle ending a session with an optional video URL.
Improved _handle_unclean_exits to ensure sessions are properly ended even if the process exits unexpectedly.
Improved LLM Tracking:

Enhanced LlmTracker to track function calls and manage events more effectively.
Improved the handling of streaming and async responses in _handle_response_v0_openai, _handle_response_v1_openai, and _handle_response_cohere methods.

Test Script
python
Copy code
import unittest
from uuid import uuid4
from agentops.client import Client
from agentops.session import Session
from agentops.helpers import get_ISO_time
from time import sleep

class TestAgentOpsEnhancements(unittest.TestCase):

    def setUp(self):
        self.client = Client(api_key="test_api_key")
        self.session_id = uuid4()

    def test_session_creation_and_expiration(self):
        session = Session(session_id=self.session_id, tags=["test_tag"])
        self.assertFalse(session.has_ended)
        self.assertFalse(session.is_expired)

        # Simulate session expiration
        session.expiration_time = get_ISO_time() - 1
        self.assertTrue(session.is_expired)

    def test_set_session_video(self):
        session = Session(session_id=self.session_id, tags=["test_tag"])
        session.set_session_video("http://example.com/video.mp4")
        self.assertEqual(session.video, "http://example.com/video.mp4")

    def test_start_and_end_session(self):
        self.client.start_session(tags=["test_start_end"])
        self.assertIsNotNone(self.client.current_session_id)

        self.client.end_session("Success", "Test end session", "http://example.com/video.mp4")
        self.assertIsNone(self.client.current_session_id)

    def test_unclean_exit_handling(self):
        session = Session(session_id=self.session_id, tags=["unclean_exit_test"])
        self.client._session = session
        self.client._worker = unittest.mock.Mock()

        # Simulate unclean exit
        self.client._handle_unclean_exits()
        atexit._run_exitfuncs()
        self.client._worker.end_session.assert_called_once()

if __name__ == "__main__":
    unittest.main()
Explanation of the Test Script
test_session_creation_and_expiration:

Creates a session and checks that it has not ended and is not expired initially.
Simulates session expiration and checks that the session is marked as expired.
test_set_session_video:

Creates a session, sets the session video URL, and verifies that the video URL is correctly set.
test_start_and_end_session:

Starts a session using the client and verifies that a session ID is set.
Ends the session and verifies that the session ID is cleared.
test_unclean_exit_handling:

Simulates an unclean exit scenario and verifies that the session end method is called to handle the unclean exit.

### DIFF
--- a/agentops/llm_tracker.py
+++ b/agentops/llm_tracker.py
@@ -13,7 +13,6 @@ import pprint
 original_create = None
 original_create_async = None
 
-
 class LlmTracker:
     SUPPORTED_APIS = {
         "litellm": {"1.3.1": ("openai_chat_completions.completion",)},
@@ -27,6 +26,9 @@ class LlmTracker:
         "cohere": {
             "5.4.0": ("chat", "chat_stream"),
         },
+        "ollama": {
+            "0.1.0": ("chat", "async_chat"),
+        }
     }
 
     def __init__(self, client):
@@ -260,8 +262,6 @@ class LlmTracker:
             StreamedChatResponse_ToolCallsGeneration,
         )
 
-        # from cohere.types.chat import ChatGenerationChunk
-
         # NOTE: Cohere only returns one message and its role will be CHATBOT which we are coercing to "assistant"
         self.llm_event = LLMEvent(init_timestamp=init_timestamp, params=kwargs)
 
@@ -375,9 +375,6 @@ class LlmTracker:
                     yield chunk
 
             return generator()
-
-        # TODO: we should record if they pass a chat.connectors, because it means they intended to call a tool
-        # Not enough to record StreamedChatResponse_ToolCallsGeneration because the tool may have not gotten called
 
         try:
             self.llm_event.returns = response.dict()
--- a/agentops/session.py
+++ b/agentops/session.py
@@ -1,7 +1,6 @@
 from .helpers import get_ISO_time
-from typing import Optional, List
+from typing import Optional, List, Dict, Any
 from uuid import UUID
-
 
 class Session:
     """
@@ -10,13 +9,15 @@ class Session:
     Args:
         session_id (UUID): The session id is used to record particular runs.
         tags (List[str], optional): Tags that can be used for grouping or sorting later. Examples could be ["GPT-4"].
+        host_env (dict, optional): Host environment details.
 
     Attributes:
         init_timestamp (float): The timestamp for when the session started, represented as seconds since the epoch.
         end_timestamp (float, optional): The timestamp for when the session ended, represented as seconds since the epoch. This is only set after end_session is called.
         end_state (str, optional): The final state of the session. Suggested: "Success", "Fail", "Indeterminate"
         end_state_reason (str, optional): The reason for ending the session.
-
+        metadata (Dict[str, Any], optional): Additional metadata for the session.
+        is_expired (bool): Flag indicating if the session has expired.
     """
 
     def __init__(
@@ -33,6 +34,8 @@ class Session:
         self.video: Optional[str] = None
         self.end_state_reason: Optional[str] = None
         self.host_env = host_env
+        self.metadata: Optional[Dict[str, Any]] = {}
+        self.is_expired: bool = False
 
     def set_session_video(self, video: str) -> None:
         """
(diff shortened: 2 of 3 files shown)


siyangqiu:
Hi @sarath59! Thanks for the contribution, but what problem does this change aim to solve? 
Enhanced LlmTracker to track function calls and manage events more effectively -> You only removed comments?
What is an expired session?
Our backend doesn't have metadata field, so it wouldn't understand it.

Was this AI generated?

sarath59:
> Hi @sarath59! Thanks for the contribution, but what problem does this change aim to solve? Enhanced LlmTracker to track function calls and manage events more effectively -> You only removed comments? What is an expired session? Our backend doesn't have metadata field, so it wouldn't understand it.
> 
> Was this AI generated?

Hey @siyangqiu , Thanks for reviewing my contribution. My motive was to make minor changes to session management and not to make huge changes. I just want to start with this for contributing. 

Coming to the expired session the goal was to add functionality to detect and handle these expired sessions to ensure they are managed appropriately and do not cause issues in the system.

I may have pushed the wrong file but my intention was to make above changes to session management. I can push changes again with correct files. 

My idea was to start small with contributions and see how it goes. I thought there could be better session management as the whole functionality resides between the sessions. 

Coming to the AI generated question. I have drafted the above description based on my thoughts to make it professional but nothing from the code. 

Also I would love if you could review the notifications feature i proposed and let me know if you would be interested in me doing that. 

thanks

agentops/session.py
```diff
@@ -43,21 +46,52 @@ def set_session_video(self, video: str) -> None:
...
         """
         self.end_state = end_state
         self.end_state_reason = end_state_reason
         self.end_timestamp = get_ISO_time()
 
+    def expire_session(self) -> None:
+        """
+        Mark the session as expired.
+        """
+        self.is_expired = True
```
braelyn-ai, line 79: we probably dont need a separate flag for expiring. we currently check `end_state` to see if it is not None, and if not, then it must have ended. so `end_state` could just be `expired` to denote that it is expired.

agentops/session.py
```diff
@@ -43,21 +46,52 @@ def set_session_video(self, video: str) -> None:
...
         Args:
             end_state (str, optional): The final state of the session. Options: "Success", "Fail", "Indeterminate"
-            rating (str, optional): The rating for the session.
             end_state_reason (str, optional): The reason for ending the session. Provides context for why the session ended.
         """
         self.end_state = end_state
         self.end_state_reason = end_state_reason
         self.end_timestamp = get_ISO_time()
 
+    def expire_session(self) -> None:
```
braelyn-ai, line 75: I dont see anywhere in this PR that the function is being used. implementing it to automatically expire may be more effort than it's worth and would likely require a dedicated thread per session. I'm not sure if we want to make the SDK more processor intensive with this. I'm open to other options though if you have others in mind!

agentops/session.py
```diff
@@ -43,21 +46,52 @@ def set_session_video(self, video: str) -> None:
         """
         self.video = video
 
+    def add_metadata(self, key: str, value: Any) -> None:
```
braelyn-ai, line 49: what kind of metadata are you thinking this would be used for? we currently have tag functionality. i'm open to building out this feature if there's a use case, but it would require a bit of dashboard and db work as well.

agentops/llm_tracker.py
```diff
@@ -27,6 +26,9 @@ class LlmTracker:
         "cohere": {
             "5.4.0": ("chat", "chat_stream"),
         },
+        "ollama": {
```
braelyn-ai, line 29: ollama support was added in another PR! can you revert this file? we'd rather keep the comments in for now

braelyn-ai:
> Implemented the is_expired property in the Session class to check if a session has expired based on the current time.

i may be blind, but i dont see where this code is in this PR

braelyn-ai:
thank you for your contribution!! i think there may be some more work here needed for this PR to be ready to merge. we also need to discuss a bit more on the metadata feature and decide if that's something we want to support.

I've left some comments on the code if you want to respond to those :)

braelyn-ai:
Closing due to inactivity

Issue #256, opened 2024-06-14
instrument_llm_calls=False + stop_instrumenting() will raise on missing llm_tracker


mbarnathan:
## 🐛 Bug Report

**🔎 Describe the Bug**
If you start up an agentops client with instrument_llm_calls false and stop_instrumenting() gets called, whether implicitly via Crew/Autogen or explicitly by the caller, client.py will try to reference an undefined self.llm_tracker member and throw. (Even though it is checked for truthiness, only getattr/hasattr will work when the value isn't set at all)

The fix is to explicitly set llm_tracker to None in the constructor, which will make the check later work properly with a falsy value. I will submit a fix.

heethjain21:
Was taking a look at this issue, and it seems like it has been already resolved.


When the PR was created https://github.com/AgentOps-AI/agentops/pull/257, the code was
```
def stop_instrumenting(self):
    if self.llm_tracker:
        self.llm_tracker.stop_instrumenting()
```

However, now the [current code](https://github.com/AgentOps-AI/agentops/blob/39958340fa765d8671dda2b5a7053e49435902cf/agentops/client.py#L362) now has check with `is not None`:
```
def stop_instrumenting(self):
    if self._llm_tracker is not None:
        self._llm_tracker.stop_instrumenting()
```

Can anyone please confirm whether there is still more to solving this issue apart from this or has it been resolved?

HowieG:
Thanks @mbarnathan for reporting and @heethjain21 for following up.

@siyangqiu says this should be a non-issue after our refactor. Tagging @bboynton97 to double check.
https://github.com/AgentOps-AI/agentops/pull/257#issuecomment-2290399028

Though Braelyn did say that her PR was not related to this one
https://github.com/AgentOps-AI/agentops/pull/368#issue-2476756793

braelyn-ai:
This should be resolved by that refactor! I'm marking this issue as closed, thank you everyone for contributing 🫶

PR #257, opened 2024-06-14
Initialize Client.llm_tracker on all codepaths.

mbarnathan:
Prevents an issue with the check in stop_instrumenting().
To properly test this, I also needed to add a parameter to override the behavior of singleton.

## 📥 Pull Request

**📘 Description**
Initialize llm_tracker to None when instrument_llm_calls is False.
Add "allow_multiple_instances" kwargs to suppress singleton decorator so the client may be tested.

#256 

**🎯 Goal**
Fix an issue that may occur when Crew and Autogen packages exist but you aren't using them.
(In my case, it popped up in a secondary file's __ main __ section)

**🧪 Testing**
Added a unit test; passed on tox.

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -98,6 +98,7 @@ class Client(metaclass=MetaClient):
             os.environ.get("AGENTOPS_ENV_DATA_OPT_OUT", "False").lower() == "true"
         )
 
+        self.llm_tracker = None
         self.config = None
 
         try:
--- a/agentops/helpers.py
+++ b/agentops/helpers.py
@@ -23,7 +23,8 @@ def singleton(class_):
     instances = {}
 
     def getinstance(*args, **kwargs):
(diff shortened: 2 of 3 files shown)


areibman:
This is awesome-- Adding @bboynton97 as a reviewer, as there might be some conflicts with the new session management strategy

siyangqiu:
Thanks for making this PR, but due to a number of updates to AgentOps, I think this PR is no longer necessary to achieve what you are looking for:

LlmTracker is only disabled by default if you have `crewai` or `autogen` in `sys.modules` rather than simply installed on your system. Furthermore, the `instrument_llm_calls` parameter of `agentops.init()` should now have precedence over the default disable behavior, so you can always just set that option to `True` regardless of packages being used (though it may still result in double logging of LLM calls)

PR #261, opened 2024-06-19
added gifs and banner images

areibman:
## 📥 Pull Request

**📘 Description**
Updated screenshots, logos, and gifs. Will be used to update other notebooks across the web


braelyn-ai:
these are cool! do we want to add them to the README as well?

braelyn-ai:
I'm not sure that we want to include these in this repo though. it adds 20MB to the repo size. we may be better off hosting in a bucket or having a separate repo for assets

areibman:
> I'm not sure that we want to include these in this repo though. it adds 20MB to the repo size. we may be better off hosting in a bucket or having a separate repo for assets

If they're going to be in docs anyway, I think it makes sense to just host it on Github which is free, has versioning, and is an effective CDN. I can't really name any other public docs sites that use external image hosting, and I also see a ton of repos with .mp4's in the readme

braelyn-ai:
> If they're going to be in docs anyway, I think it makes sense to just host it on Github which is free, has versioning, and is an effective CDN. I can't really name any other public docs sites that use external image hosting, and I also see a ton of repos with .mp4's in the readme

We could look into using git LFS. I believe GitHub automatically offers a certain amount of blob storage to be used with LFS. Blobs/binary files still get an immutable and versioned pointer, but the file isn't stored in the repo, rather on a CDN

braelyn-ai (approved):

PR #271, opened 2024-06-26
revert Apply @check_init to decorators

siyangqiu:
## 📥 Pull Request

**📘 Description**
Adding @check_init directly to decorators causes a more serious issue than it fixes. Reverting this for now. A more significant overhaul of SDK state will be used to fix this issue


braelyn-ai (approved):
AAINO _(appears acceptable in my opinion)_

PR #315, opened 2024-07-26
Rework initialization, improve logging, and other refactoring

siyangqiu:
## 📥 Pull Request

**📘 Description**
This change addresses a number of common pain points with the AgentOps SDK:

- The `init()` method is expected to be the first entry-point into AgentOps code. 
-- Unexpected behavior is extremely common when this is not the case - often happens because decorators in submodules are called when the submodules are imported. 
-- AgentOps integrations cannot override the default configuration as there is no `Client()` to set configuration options

- Log messages are suppressed before `init()` is called. Some warnings messages are not revealed to users as a result
- Errors messages can be confusing. The some errors are split up into multiple logs. Some errors are not descriptive enough because when the issue happens, some information is no longer available

** The Change**

- HTTP Client now throws exceptions so that the calling method can log more descriptive messages
- Improved error messages in general
- The `Client()` now has "initialized" state variable - only set to true when everything is ready when all required info is provided
- The `Client()` constructor now only sets default configuration and does not take arguments. `Client().configure()` is now used to configure the client. `init()` calls `configure()` and then initializes the client.


agentops/config.py
```diff
@@ -1,149 +1,74 @@
...
-    def __init__(
+from typing import List, Optional
+from uuid import UUID
+
+from .log_config import logger
+
+
+class Configuration:
+    def __init__(self):
+        self.api_key: Optional[str] = None
```
siyangqiu, line 9: Make all variables private

agentops/client.py
```diff
@@ -194,14 +159,26 @@ def set_tags(self, tags: List[str]) -> None:
...
-            return
+            return logger.warning(
+                "Could not set tags. Start a session by calling agentops.start_session()."
+            )
+        else:
+            session.set_tags(tags=tags)
 
-        session.set_tags(tags=tags)
+    def add_default_tags(self, tags: List[str]) -> None:
+        """
```
siyangqiu, line 164: Add a get default tags method

agentops/session.py
```diff
@@ -252,3 +332,6 @@ def wrapper(*args, **kwargs):
             return func(*args, **kwargs)
 
         return wrapper
+
+
+active_sessions: List[Session] = []
```
siyangqiu, line 334: move to client.py if possible

agentops/__init__.py
```diff
@@ -62,50 +50,95 @@ def init(
...
+    """
+    Start a new session for recording events.
+
+    Args:
+        tags (List[str], optional): Tags that can be used for grouping or sorting later.
+            e.g. ["test_run"].
+        inherited_session_id: (str, optional): Set the session ID to inherit from another client
+    """
+    if not Client().is_initialized:
+        return logger.warning(
```
siyangqiu, line 152: make this an pre-init warning

agentops/__init__.py
```diff
@@ -131,45 +172,24 @@ def end_session(
...
-
-@check_init
 def record(event: Union[Event, ErrorEvent]):
     """
     Record an event with the AgentOps service.
 
     Args:
         event (Event): The event to record.
     """
+    if Client().is_multi_session:
```
siyangqiu, line 201: Add pre init warnings

agentops/config.py
siyangqiu: Actually I don't want to create getters for everything

agentops/session.py
siyangqiu: Not possible

agentops/__init__.py
```diff
@@ -44,102 +31,111 @@ def init(
...
-
-    c = Client(
+    Client().unsuppress_logs()
+    if Client().is_initialized:
+        return logger.warning("AgentOps has already been initialized")
+
+    if tags is not None:
+        logger.warning("The 'tags' parameter is deprecated. Use 'default_tags' instead")
+        if default_tags is None:
+            default_tags = tags
```
braelyn-ai, line 70: we have tests for add_tags and set_tags rn. it might be worth writing some for default tags too, they're pretty simple

agentops/__init__.py
```diff
@@ -180,6 +219,17 @@ def add_tags(tags: List[str]):
     Args:
         tags (List[str]): The list of tags to append.
     """
+    if Client().is_multi_session:
+        return logger.warning(
```
braelyn-ai, line 223: def not right now, but we could create a logger mock and test all of these down the road

examples/agent.py
```diff
@@ -0,0 +1,11 @@
+from agentops import track_agent
+
+
+@track_agent("name")
```
braelyn-ai: do we want this file?

examples/temp.py
```diff
@@ -0,0 +1,10 @@
+import agentops
+from agentops import ActionEvent
+from agent import agent
+
```
braelyn-ai: especially this one. it appears to be a local test file. if you want to leave it in the repo, the best place is `tests/core_manual_tests`

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "agentops"
-version = "0.3.4"
+version = "0.4.0"
```
braelyn-ai: are we sure we want to make this a change to 0.4? is a big refactor internally, but to the developer it's not a significant change as there are no breaking changes or major features added. i personally think this should just be a patch but open to discussion

braelyn-ai (approved):
nits but nothing blocking

Issue #291, opened 2024-07-09
'Client' object has no attribute 'llm_tracker'


rmico77:
>>> # Criando as LLMs
>>> llama8=ChatGroq(model_name="llama3-8b-8192", api_key=GROQ_API_KEY)
>>> llama70=ChatGroq(model_name="llama3-70b-8192", api_key=GROQ_API_KEY)
>>> gpt3=ChatOpenAI(model="gpt-3.5-turbo", api_key=OPENAI_API_KEY)
>>>
>>>
>>> # Ferramenta de busca em PDF
>>> search_tool = PDFSearchTool(pdf=r'C:\Users\rmico\Desktop\CrewAI\Gilbert_Developmental_Biology_2019.pdf')
>>>
>>> # Agente de Leitura
>>> leitor = Agent(
...     role='Leitor Acadêmico',
...     goal='Extrair informações específicas do livro acadêmico em PDF',
...     backstory="Você é um leitor experiente em textos acadêmicos.",
...     tools=[search_tool],
...     verbose=True,
...     allow_delegation=False,
... )
🖇 AgentOps: Error: 'Client' object has no attribute 'llm_tracker'
2024-07-09 17:22:38,859 - 13680 - meta_client.py-meta_client:51 - WARNING: Error: 'Client' object has no attribute 'llm_tracker'
🖇 AgentOps: Failed to track an agent. This often happens if agentops.init() was not called before i nitializing an agent with the @track_agent decorator.
2024-07-09 17:22:38,908 - 13680 - agent.py-agent:23 - WARNING: Failed to track an agent. This often happens if agentops.init() was not called before initializing an agent with the @track_agent decorator.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Users\rmico\AppData\Local\Programs\Python\Python312\Lib\site-packages\agentops\agent.py", line 25, in new_init
    logger.warning(
    ^^^^^^^
  File "C:\Users\rmico\AppData\Local\Programs\Python\Python312\Lib\site-packages\agentops\agent.py", line 19, in new_init
    original_init(self, *args, **kwargs)
  File "C:\Users\rmico\AppData\Local\Programs\Python\Python312\Lib\site-packages\crewai\agent.py", line 102, in __init__
    super().__init__(**config, **data)
  File "C:\Users\rmico\AppData\Local\Programs\Python\Python312\Lib\site-packages\crewai\agents\agent_builder\base_agent.py", line 128, in __init__
[...]

PR #320, opened 2024-07-27
bugfix: 291 - llm_tracker is not an attribute in client class

andkoma:
📘 Description
initialize the llm_tracker in class client with None

🔄 Related Issue (if applicable)
https://github.com/AgentOps-AI/agentops/issues/291

🎯 Goal
fix the bug, that the client class has in some cases no attribute for llm_tracker.

🔍 Additional Context
Looks like the class init is not complete processed and the attribute did not been created in some circumstances?

🧪 Testing
Fix in local environment, re-run the crewai application with agentops in two modes.
a) enabled
b) disabled.

### DIFF
--- a/agentops/client.py
+++ b/agentops/client.py
@@ -95,8 +95,9 @@ class Client(metaclass=MetaClient):
             os.environ.get("AGENTOPS_ENV_DATA_OPT_OUT", "False").lower() == "true"
         )
 
+        self.llm_tracker = None
+        
         self.config = None
-
         try:
             self.config = ClientConfiguration(
                 api_key=api_key,


areibman:
Thanks for this! @bboynton97 what do you think?

braelyn-ai:
Thank you for the contribution @andkoma! I think there's an underlying issue here we need to investigate, but this does prevent the case for an exception halting the execution of an agent due to AgentOps.

@areibman I say we merge this and investigate the issue after

braelyn-ai:
@andkoma Could you merge upstream and run `black .` 

If you dont have Black installed, run:
```
pip install black 'black[jupyter]'
```

andkoma:
Ok. 
My change was just the insert of a single line statement.
This could not be the root cause for the merge result.

Regards

braelyn-ai:
There's no merge conflict. Your branch is behind and not formatted properly

braelyn-ai (approved):

andkoma:
I just checked this behavior,
I’ve made the change inside the web interface of the forked repo.
Maybe in this case the embedded editor destroys your expected formatting.

I will drop the fork and create a fresh one.

Regards and sorry for any circumstances
Andrew

PR #328, opened 2024-08-02
Adding HTTPClient get and updating timetravel to use it

HowieG:
## 📥 Pull Request

**📘 Description**
We only had post, now we have get

ENG-305

**🎯 Goal**
v2/get_ttd route in api is a get so we need to support this in sdk

**🔍 Additional Context**
Any extra information or context to help us understand the change?

**🧪 Testing**
tests/core_manual_tests/time_travel.py

### DIFF
--- a/agentops/http_client.py
+++ b/agentops/http_client.py
@@ -119,3 +119,58 @@ class HttpClient:
             raise ApiServerException("API server: - internal server error")
 
         return result
+
+    @staticmethod
+    def get(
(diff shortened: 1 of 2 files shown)


HowieG:
This is mostly a duplicate of post() so lmk if we want to modularize it more however I usually see these things defined independently even if it's copypaste

braelyn-ai (approved):

PR #334, opened 2024-08-05
fix: speed up version check

braelyn-ai:
The agentops version check was halting execution of other code and using the pip CLI.
Switching to the PyPi API and running on a separate thread fixes this.


siyangqiu (approved):
Nice

PR #346, opened 2024-08-09
separated llm tracker by model

braelyn-ai:
Each LLM now has functions in a different file. I will probably continue this to use inheritance and consistent function names


agentops/llms/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+                    # Patch openai v1.0.0+ methods
+                    if hasattr(module, "__version__"):
+                        module_version = parse(module.__version__)
+                        if module_version >= parse("1.0.0"):
+                            provider = OpenAiProvider(self.client)
+                            provider.override()
+                        else:
+                            raise DeprecationWarning(
+                                "OpenAI versions < 0.1 are no longer supported by AgentOps. Please upgrade OpenAI or "
+                                "downgrade AgentOps."
```
siyangqiu: Can we specify 0.3.9 here if this is 0.3.10 or 0.3.x if this will be 0.4? That way users know which version to use.

agentops/llms/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+from .litellm import LiteLLMProvider
+from .ollama import OllamaProvider
+from .openai import OpenAiProvider
+
+original_func = {}
+original_create = None
+original_create_async = None
+
+
+class LlmTracker:
```
siyangqiu, line 22: Just a thought: Do we want to call this instrumenter instead? Or we can call the folder instrumenter. In theory, we can instrument packages besides just LLMs (ie vectordbs)

agentops/llms/__init__.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def stop_instrumenting(self):
+        openai_provider = OpenAiProvider(self.client)
+        openai_provider.undo_override()
+
+        groq_provider = GroqProvider(self.client)
+        groq_provider.undo_override()
+
+        cohere_provider = CohereProvider(self.client)
+        cohere_provider.undo_override()
```
siyangqiu: What about litellm and ollama?

siyangqiu (approved):

siyangqiu:
I ran a core manual test and it worked

agentops/llms/__init__.py
braelyn-ai: Okay, ill add these and change the manual tests to utilize it

agentops/llms/__init__.py
braelyn-ai: I'd rather leave this pattern as LLM stuff only and we can create other implementations for other instrumenting patterns

PR #352, opened 2024-08-13
fix rlock issue

siyangqiu:
## 📥 Pull Request

**📘 Description**
Deepcopy fails when there are non-pickleable objects in the "queue". Because events often contain "self" of some user created objects, we cannot be sure there are no non-pickleable objects. Simple copy should be sufficient here

### DIFF
--- a/agentops/session.py
+++ b/agentops/session.py
@@ -268,7 +268,7 @@ class Session:
         if not self.is_running:
             return
         with self.lock:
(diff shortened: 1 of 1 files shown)


siyangqiu:
@bboynton97 Do you remember if there was an autogen or particular agent that was double recording or something?

braelyn-ai (approved):

PR #355, opened 2024-08-14
Modify test-notebooks.yml to run manually and on push to main

HowieG:
## 📥 Pull Request
Resolves ENG-594

TODO: Will make it block merges to main once we fix all the notebooks

**🎯 Goal**
Not run the notebooks on every commit to a PR

### DIFF
--- a/.github/workflows/test-notebooks.yml
+++ b/.github/workflows/test-notebooks.yml
@@ -1,9 +1,6 @@
 name: Test Notebooks
 on:
-  push:
-    branches: [ main ]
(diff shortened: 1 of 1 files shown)


HowieG:
@siyangqiu This exposes a button: Actions > Test Notebooks > Run workflow > Run workflow
And it runs on push to main but it won't yet block merges to main. That's done in branch protection settings but we have to fix all the notebooks first otherwise all merging to main will be blocked

braelyn-ai (approved):

PR #367, opened 2024-08-20
event decorators have default names

braelyn-ai:
@record_action and @record_tool both will use the function's name as the default event name if none is supplied.

Tests


braelyn-ai:
Closes ENG-584

HowieG (approved):

PR #381, opened 2024-09-09
Capture track agent

braelyn-ai:
If a user initializes an agent with a `@track_agent` decorator, it not gets added to a queue rather than erroring out.

closes ENG-620


agentops/decorators.py
```diff
@@ -344,14 +338,7 @@ def new_init(self, *args, **kwargs):
...
                 except AttributeError as e:
-                    Client().add_pre_init_warning(
-                        f"Failed to track an agent {name} because agentops.init() was not "
-                        + "called before initializing the agent with the @track_agent decorator."
-                    )
-                    logger.warning(
-                        "Failed to track an agent. This often happens if agentops.init() was not "
-                        "called before initializing an agent with the @track_agent decorator."
-                    )
+                    print(e)
```
siyangqiu: We should still use the logger and pre_init_warning system. This piece could be used inside of a library like crew and cause printing to happen without agentops.init()

siyangqiu (approved):

agentops/decorators.py
braelyn-ai: you're right. i changed the wording so it's still applicable

PR #388, opened 2024-09-14
Fixed Time Travel output for multithreading and clarity

HowieG:


HowieG:
TODO @areibman wants the print statements to be prefixed so we don't repeat ourselves

HowieG:
> TODO @areibman wants the print statements to be prefixed so we don't repeat ourselves

Done

siyangqiu (approved):

Issue #380, opened 2024-09-08
Add duration + LLM analytics at end of session in CLI


areibman:
## 🚀 Feature Request

**💡 Got a brilliant idea?**
Right now, we return total session cost: `🖇 AgentOps: This run's cost $0.00`

Would be awesome if we also show:
* Number of LLM calls
* Number of tool calls
* Total duration of session

heethjain21:
Created a PR: https://github.com/AgentOps-AI/agentops/pull/392

Can add more metrics if needed (for now, just added these 3 additional metrics)

PR #387, opened 2024-09-13
feat: disable AgentOps flag

heethjain21:
## 📥 Pull Request

**📘 Description**
- Added `is_disabled` flag during `init()` which is further passed to the Client config
- Added global `enable()` and `disable()` functions to perform the same
- When `is_disabled` is set to True, all sessions, event recording are disabled
- Updated docs for the same

**🔄 Related Issue (if applicable)**
[Create a way to disable AgentOps functions entirely #349
](https://github.com/AgentOps-AI/agentops/issues/349)

**🎯 Goal**
- Add a global way to disable AgentOps completely with a flag

**🔍 Additional Context**
- None

**🧪 Testing**
- Added test cases for `sessions`, `canary` (wasn't sure if it was needed here), `record_actions`, `record_tools`

PR #392, opened 2024-09-17
feat: session end analytics: tools, llms, duration

heethjain21:
## 📥 Pull Request

**📘 Description**
- Counting the no of events based on event type
- Logging the no of events on session end
- Also logging the session duration

[Add duration + LLM analytics at end of session in CLI #380
](https://github.com/AgentOps-AI/agentops/issues/380)

**🎯 Goal**
- Have some additional metrics shown on session ending

**🔍 Additional Context**
- I checked if there was an API for GET session by Id, but couldn't find it, so the next best solution was to just simply calculate the events

**🧪 Testing**
- Not needed, since its just logging


HowieG:
Hey @heethjain21 - we're able to calcuate these API-side and for duration we do have that in our dashboard as one of our key features - https://app.agentops.ai/
So we don't have a need for this PR unfortunately

HowieG:
I guess the intention was to log them at the end just like we log the cost. Again not sure if there's a need and I think we have to keep the terminal more or less clean because developers that use our sdk may not want a lot of noise especially considering we're an add-on for observability not like a primary tool for accomplishing a task

HowieG:
Actually I think duration is valuable! Asking team about counts

HowieG (approved):

heethjain21:
Hi @HowieG , 

This PR's workflow didn't run all the tests (earlier), so couldn't realize `dateutil` package messed up.
I had it installed the package locally, so couldn't replicate it for the tests locally (and hence couldn't catch the tests failure)

I have corrected the issue in this Pull Request: https://github.com/AgentOps-AI/agentops/pull/387
Please take a look, and merge it so the PR workflows get fixed.

areibman:
Thanks @heethjain21 !

HowieG:
Hey @heethjain21 thanks for followup my bad I also merged without checking the workflows. I ended up replacing the dateutil logic with the builtin datetime to minimize addition of new dependencies

heethjain21:
Alright, thanks!

PR #435, opened 2024-09-30
make start_session non blocking

siyangqiu:
## 📥 Pull Request

**📘 Description**
Make start session non-blocking


areibman:
I branched off of this for a separate PR that merged. We'll solve the start_session blocking separately with a different implementation

PR #441, opened 2024-10-14
Update session.py to fix local variable 'formatted_cost' when it is not associated with a value

knivore:
Format formatted_cost to be outside of the token_cost if else condition.

## 📥 Pull Request

**📘 Description**
Format formatted_cost block of code to shift outside of the token_cost if else condition due to the issue when there are no token_cost, formatted_cost variable is not associated with a value

**🧪 Testing**

### DIFF
--- a/agentops/session.py
+++ b/agentops/session.py
@@ -142,13 +142,14 @@ class Session:
             token_cost_d = Decimal(0)
         else:
             token_cost_d = Decimal(token_cost)
-            formatted_cost = (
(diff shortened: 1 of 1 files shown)


areibman (approved):
Good catch!

areibman:
Brilliant-- thanks @knivore! Merging, we'll try to stage a release later today

HowieG:
Note: we might have expected explicitly "0" on our api server and this change now makes it 0.00

PR #472, opened 2024-10-30
Make end_state default to "Indeterminate" when creating session

HowieG:
We had many sessions with null end_states which was causing our pie chart to seem to be "missing" sessions


areibman (approved):
Fine with this. We made an early on decision to have it be none but TBH makes sense to have it default

PR #474, opened 2024-10-31
[do not merge] migrate build system to hatchling

### LINKED ISSUE #473: [Bug]: Import resolution breaks with --editable on `uv`, `rye` packaage managers
### Contact Details

teocns@gmail.com

### 📦 Package Version

latest

### 🎞️ Framework Version

N/A

### 🔎 Describe the Bug

The issue occurs when installing the package in editable mode using either `uv` or `rye`, with a virtual environment activated, and only affects the LSP only: both Pyright and Pylance(vscode) produce the same effect. 

Things tried unsuccessfully:

- Using a different setuptools version 
- A similar [issue](https://github.com/microsoft/pyright/issues/8367) suggested a fix happened in the latest Pyright release, though it did not seem to address this issue in particular
- Manipulating import paths
- Numerous `[tool.setuptools]` configurations
- ... 

What's going on under the hood though?
Installing a package in editable mode creates a link under `site-packages` , with a file named something similar to `__editable__.agentops-0.3.14.finder.__path_hook__`, which contains the absolute path to the package that has just been installed as `--editable`.  It is also a hook, as the name suggests, that would eventually take care of modifying your `PYTHONPATH`

Since I was able to get a type hint with that filename (rather than `agentops`), it  suggests the hook did not run at installation, for whatever unknown reason that I did not consider investigating further.

`setuptools` may be hard to get straight since it's the lowest level of a package builder you get. I personally try to stay away from it to avoid problems, as it takes time and patience to get straight especially for the use case we're dealing with here.

What solved the issue instantly without further changes was [switching to hatchling](https://github.com/AgentOps-AI/agentops/commit/b60e653a4c616405a9bf67862684695cce202fbd), a modern build system adopted across `astral.sh` tools and that I personally use across python projects.

https://github.com/AgentOps-AI/agentops/commit/b60e653a4c616405a9bf67862684695cce202fbd#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R1-R3

### 🤝 Contribution

- [X] Yes, I'd be happy to submit a pull request with these changes.

teocns:
Fixes #473 

---

### Why Hatch?
[Why hatch?](https://hatch.pypa.io/latest/why/)

### Who Uses Hatch?
- [Quivr](https://github.com/QuivrHQ/quivr/blob/main/core/pyproject.toml)

### Migrating from Setuptools to Hatch: Is It Safe?
- Yes, Hatchling is compatible with modern PEP standards (e.g., PEP 517 and 518).

### Does It Require Me to Install Extra Dependencies?
- No. `pip` automatically resolves Hatchling in a temporary environment for the build if isolated builds are enabled (typically the default). This is similar behavior to setuptools and other PEP-compliant build backends specified in `pyproject.toml`.

### When Hatch May Not Be Ideal
- Hatch may not be ideal for projects involving complex builds, such as those compiling C or other extension modules, due to limited support for handling compiled dependencies. (Not relevant for this project)

### Before Merging: Feedback Loop
Let's keep this PR open for feedback from peers to measure its broader impact. Although several tests have been conducted, ensuring that #473 isn’t an issue unique to one setup is essential. Otherwise, we should consider if investing more time into troubleshooting `setuptools` is worth it.

---

### TODO Tasks:
- [ ] Review and gather feedback from peers @areibman @siyangqiu 
- [ ] Confirm compatibility with CI (e.g., Tox).
- [ ] Validate the impact across different environments.
- [ ] Ensure no critical build failures or edge cases emerge.
- [ ] Document any specific configuration changes needed for CI integration.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,10 +1,11 @@
 [build-system]
-requires = ["setuptools>=61.0", "wheel"]
-build-backend = "setuptools.build_meta"
+requires = ["hatchling"]
+build-backend = "hatchling.build"
 
 [project]
 name = "agentops"
 version = "0.3.14"
+
 authors = [
   { name="Alex Reibman", email="areibman@gmail.com" },
   { name="Shawn Qiu", email="siyangqiu@gmail.com" },
@@ -45,3 +46,9 @@ max_line_length = 120
 
 [project.scripts]
 agentops = "agentops.cli:main"
+
+[tool.hatch.metadata]
+allow-direct-references = true
+
+[tool.hatch.build.targets.wheel]
+packages = ["agentops"]


areibman:
Closing this for now-- `uv` is surpassing `rye`, so this will not be the most valuable contribution for the time being

teocns:
This had me suffering, but here goes the solution

https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior

https://github.com/astral-sh/uv/issues/1708

... in the middle of nowhere.

PR #476, opened 2024-10-31
liteLLM bug 

sarath59:
## 📥 Pull Request

**📘 Description**

The problem occurs because:
1. When both libraries are imported, the override methods are competing for the same OpenAI completion methods
2. LiteLLM internally uses OpenAI's format, and when both are imported, the overrides are interfering
3. The global variables storing original methods are being overwritten

**🧪 Testing**

Test Coverage
1. Synchronous and asynchronous API calls
2. Streaming and non-streaming responses
3. Method override and restoration
4. Error handling and logging
5. Multiple consecutive API calls
6. Integration with LiteLLM present

Validation Results
1. Successfully tracked OpenAI API calls with LiteLLM imported
2. Properly handled method overrides and restorations
3. Maintained compatibility with both sync and async operations
4. Verified proper error handling and logging
5. Confirmed session management and cleanup
Screenshot 2024-10-31 at 11 16 41 AM
Screenshot 2024-10-31 at 11 17 28 AM

PR #480, opened 2024-10-31
fix: litellm conflict with openai on lib override

### LINKED ISSUE #471: [Bug]: LiteLLM conflicts with OpenAI
### Contact Details

_No response_

### 📦 Package Version

0.3.14

### 🎞️ Framework Version

openai==1.52.2, litellm==1.50.2

### 🔎 Describe the Bug

LiteLLM alone: Works
image

OpenAI alone: Works
image

Import both OpenAI and LiteLLM - Doesn't Work
image

Same does not apply when you only use LiteLLM
image

### 🤝 Contribution

teocns:
Fixes #471 

- Removed `original_oai_create` and `original_oai_create_async` variables since we no longer override OpenAI's methods
- Modified `override()` and `undo_override()` to only handle LiteLLM's methods 
- Updated `_override_completion()` and `_override_async_completion()` to only store and patch LiteLLM's methods

This way, when both providers are used:
- OpenAIProvider will handle overriding OpenAI's completion methods 
- LiteLLMProvider will only handle overriding LiteLLM's completion methods 
- No more conflicts between the two providers ✅ 

### A bit more of explanation

LiteLLM's completion method is completely separate from OpenAI's `Completions.create`
Even though LiteLLM uses OpenAI's format internally, it has its own implementation
When we call `litellm.completion()`, it doesn't actually call OpenAI's `Completions.create`

##### Breakdown 

- `test_provider_override_independence()`:
  1. Verifies that each provider only overrides its own methods
  2. Checks that the original methods are properly stored and restored
  3. Ensures no cross-contamination between providers
- `test_provider_override_order_independence()`:
  1. Verifies that the order of provider initialization doesn't matter
  2. Tests both possible orders (OpenAI first vs LiteLLM first)
  3. Ensures each provider's overrides remain intact regardless of order
  
##### What's being verified
- Each provider only overrides its intended methods
- Original methods are properly stored and can be restored
- The order of provider initialization doesn't affect the overrides
- No provider interferes with another provider's overrides

```
python tests/test_providers.py
```

### DIFF
--- a/agentops/llms/litellm.py
+++ b/agentops/llms/litellm.py
@@ -25,20 +25,11 @@ class LiteLLMProvider(InstrumentedProvider):
         self._override_completion()
 
     def undo_override(self):
-        if (
-            self.original_create is not None
-            and self.original_create_async is not None
-            and self.original_oai_create is not None
-            and self.original_oai_create_async is not None
-        ):
+        if self.original_create is not None and self.original_create_async is not None:
             import litellm
-            from openai.resources.chat import completions
-
+            
             litellm.acompletion = self.original_create_async
             litellm.completion = self.original_create
-
-            completions.Completions.create = self.original_oai_create
-            completions.AsyncCompletions.create = self.original_oai_create_async
 
     def handle_response(
         self, response, kwargs, init_timestamp, session: Optional[Session] = None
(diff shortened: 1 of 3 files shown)


areibman:
I ran this in `python` and it didn't work for OpenAI

```
import litellm
import openai
import agentops

agentops.init(default_tags=["litellm"])

message = [{"role": "user", "content": "Write a 12 word poem about secret agents."}]

litellm.completion(model="claude-3-sonnet-20240229", messages=message, temperature=0)

client = openai.OpenAI()

client.chat.completions.create(model="gpt-4o", messages=message, temperature=0)```

areibman:
@the-praxs Can you give it a look?

dot-agi:
> @the-praxs Can you give it a look?

Testing it - same issue as yesterday - only the first `create` uses the patched function.

I am trying to troubleshoot the root cause.

teocns:
closing in favor of #476

areibman:
If we want to avoid client conflicts, perhaps we can instead use their official observability documentation https://docs.litellm.ai/docs/observability/arize_integration

teocns:
Since the other PR was closed by the author will consider this as "Resolves #471 "

Issue #499, opened 2024-11-10
[Docs]: Anthropic Tool Usage Documentation Fix


TKTSWalker:
### 📘 Current State of Documentation

While working on the documentation for Anthropic, I noticed a few things;

1. There seems to be trouble getting the text as a stream and it instead likes to get the output as a singular item.
2. The base AgentOps reference puts everything into a string but this is wrong; Anthropic uses a block system similar to JSON
3. Multiple blocks may be within, with one being for the AI's text output and another for the tool start
4. Understanding how to use the tools may be one of the hardest things for a newcomer, given it does require understanding context and bringing in the text 1 to 1 may not always be a good idea.

### 📖 Suggested Improvement

I am currently working on an additional journal which

1. Understands the context from the chunk output, showing users how to extract the tool, inputs and AI dialogue
2. Shows how to use the tools in a simple way, while also showing different kinds of output methods
3. Shows the user how to add the context from the tool into the LLM's memory to receive an answer

This should take me about two days to complete, given i'm looking at either doing Ollama's documentation or continuing putting effort towards the Gemini issue!

### 🔗 Affected Documentation Pages

https://github.com/AgentOps-AI/agentops/blob/main/examples/anthropic_examples/anthropic_example.ipynb

https://github.com/AgentOps-AI/agentops/blob/main/examples/anthropic_examples/anthropicguide.md

### 🤝 Contribution

- [X] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the Agentops team to handle this update.

dot-agi:
Assigning you this issue, address it when you open a PR. Thank you so much!

TKTSWalker:
Hey, I tried pointing to this issue but i'm not sure why it wouldn't work! I made a PR request!

PR #503, opened 2024-11-11
Added More Indepth Tool Documentation For Anthropic API

TKTSWalker:
## 📥 Pull Request

**📘 Description**

1. Made an example showcasing simple tool calling while showing a second which showcased both using multiple calls at once and requesting an input
2. Left code blocks which could be used for listening for a singular tool or multiple
3. Showed how to initialize the tools and give the data to the LLM as context
4. Reflected as such within the Anthropic general guide journal 

This also handles the problem high lighted by #499!

**🧪 Testing**

The major outputs are still visible with their results intact within the journal!

Also on the journal for the barebones reference, it might be a good idea to remove the tool segment and point to the document! Finally, it might be good to make a third link on the readme page with a link to the journal itself although you may already have plans here!


TKTSWalker:
@the-praxs All done! The main file referenced is [here!](https://github.com/TKTSWalker/agentops/blob/main)

TKTSWalker:
Realized the problem didn't properly link issue 

Fix bug #499

TKTSWalker:
#499

TKTSWalker:
Fix bug https://github.com/AgentOps-AI/agentops/issues/499

TKTSWalker:
Fix bug #499

examples/anthropic_examples/agentops-anthropic-understanding-tools.ipynb
areibman, line 1: image
Can we make this cell smaller? The example, although creative, is a bit too long to illustrate the point we're demonstrating

areibman, line 1: Minor idiomatic python nit:
```
print (tool_use_count)
```
No space between print and `(`

```

while loop < tool_use_count: #We will get the tools now
    tool_use_block = response.content[loop + 1] #We start at 1 since 0 holds the AI mesage
    tool_name = tool_use_block.name
    tool_input = tool_use_block.input
    
    if tool_name == "inventoryscan_tool":
        # Call the inventoryscan function for inventoryscan_tool
        invscanres = inventoryscan()
    elif tool_name == "enemyscan_tool":
        # Get the amount for enemyscan_tool
        amount = tool_input['amount']
        # Call the enemyscan function with the amount
        enemyscanres = enemyscan(amount)
        
    loop = loop + 1 
```

This is a pretty uncommon way to loop in python. I'd recommend a for loop:
```
for _ in range(tool_use_count):
    ...
```


```
invscanres = ""
enemyscanres = ""
```
snake_case for readability 
```
inv_scan_res = ""
enemy_scan_res = ""
initial_messages
```

areibman, line 1: image
Extra markdown cell not deleted

areibman, line 1: Print the content `response.content`
image

areibman: TitleCase is reserved for classes in python
```
CombatCasualties 
Missions 
```
```
combat_casualties
missions 
```

areibman (changes requested):
Handful of changes. Also can we rename the notebook or combine it with another one? We have several example notebooks in this examples directory that more or less do the same things

examples/anthropic_examples/agentops-anthropic-understanding-tools.ipynb
TKTSWalker: Hello, many apologies! I'll keep the below in mind when making more journals in the future! I would suggest renaming the notebook rather than combining it with another, the reason mainly being the ways in which the tools are extracted! In specific, the system loops through each "block" in order to get the appropriate data in relation to the tool that should be used!

Shortening the example, I went overkill admittedly!

TKTSWalker: I purposefully didn't use the for loop as when originally using it, I had an error related to the first content block (Which from what I can tell is always the AI dialogue), so instead I had to work outside the box!
Basically we get the amount of times the word ToolBlock appears right (Because it will always appear just once unless the user does something that has it appear many times)! Then if we start loop on 0, we can just add a value of 1 (To account for the first item basically not being readable)! Then as soon as loop is greater than the amount of times the item appears we know that we are out of range!
Thinking about it, it could technically be possible to turn each chunk into a sort of string list and then remove the first chunk like that but i'm not %100 sure if it'd be worth it (I'm still willing to if you believe it is)!

Fixing variables, that is something else I should've fixed!

Same with TitleCase, a bad habit from my C# naming shenanigans

TKTSWalker: Deleting extra markdown cell!

TKTSWalker: It will be fixed!

TKTSWalker (commented):
Good to go!

TKTSWalker:
Sorry for the small time gap, was preoccupied with my own project for a little! I've updated the journal, if there are any more problems please feel free to let me know!

areibman (approved):

areibman:
Thanks for the PR @TKTSWalker !

TKTSWalker:
No problem, looking forward to doing more!

Issue #500, opened 2024-11-11
[Feature]: TaskWeaver Support


dot-agi:
### 💡 Feature Description and Proposed Solution

https://github.com/microsoft/taskweaver/

Add support for TaskWeaver, similar to AutoGen.

### 🤝 Contribution

- [ ] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the AgentOps team to handle this update.

PR #506, opened 2024-11-14
agentops_property - internal objects tracking the right way

teocns:
Introduces the  `agentops_property` descriptor for safe and consistent objects tracking;

- The `check_call_stack_for_agent_id` function has been refactored to utilize this property descriptor improving the process of identifying agent IDs from the call stack. 

- Property access nomenclature is standardized to `_agentops_<name>` for _private attrs_ and `agentops_<name>` for _public attrs_


agentops/decorators.py
```diff
@@ -311,30 +312,36 @@ def sync_wrapper(*args, session: Optional[Session] = None, **kwargs):
...
+                        self._agentops_session_id = session.session_id
 
                     Client().create_agent(
-                        name=self.agent_ops_agent_name,
-                        agent_id=self.agent_ops_agent_id,
+                        name=self._agentops_agent_name,
+                        agent_id=self._agentops_agent_id,
                         session=session,
                     )
                 except AttributeError as e:
```
areibman: I'm not sure what this bot is talking about lol

agentops/descriptor.py
areibman, line 1: I'm not entirely sure I understand the purpose of this file. Why are we replacing the helper function with `AgentOpsDescriptor`?

agentops/helpers.py
```diff
@@ -99,22 +106,7 @@ def remove_unwanted_items(value):
...
-            if hasattr(var, "agent_ops_agent_id") and getattr(
-                var, "agent_ops_agent_id"
-            ):
-                logger.debug(
-                    "LLM call from agent named: %s",
-                    getattr(var, "agent_ops_agent_name"),
-                )
-                return getattr(var, "agent_ops_agent_id")
-    return None
+    return AgentOpsDescriptor.from_stack()
```
areibman: Everywhere in this PR we instantiate `AgentOpsDescriptor` inline instead of having class instance variables. Why are we doing this instead of just running `from_stack` as a separately defined helper function?

areibman (dismissed):

agentops/helpers.py
teocns: The signature would've better been off `AgentOpsDescriptor.get_from_stack(property="agent_id")`

Reason you want a descriptor is because you need a distinguishable way of identifying your own properties (agentops props)

> Everywhere in this PR we instantiate AgentOpsDescriptor inline instead

...it's not being instantiated, it's a static method being invoked, as if it was a function under a module named `AgentOpsDescriptor`. 

Most of the "why" are answered with behavior encapsulation

teocns:
The previous helper was just a helper in the middle of a `helpers` utility file. This introduces a new concept that's quite fundamental for what's going on in the AgentOps library: mocking / patching, property lookup, etc. Challenge being faced right now is tracking of _AgentOps'_ own properties installed onto objects. 

For instance, say you wanna lookup `_agentops_agent_id` onto a property.

```

class Lies:

    def __init__(self):
        self._agentops_agent_id = uuid4()
      

    def __getattr__(self):
      return "lied"



Lies()._agentops_agent_id # lied
```

Without a descriptor, there is no way to determine whether the value accessed at `_agentops_agent_id` is genuine, which is exactly what was happening with retrieving the agent_id from call stack: a sea of lies.

areibman:
> The previous helper was just a helper in the middle of a `helpers` utility file. This introduces a new concept that's quite fundamental for what's going on in the AgentOps library: mocking / patching, property lookup, etc. Challenge being faced right now is tracking of _AgentOps'_ own properties installed onto objects.
> 
> For instance, say you wanna lookup `_agentops_agent_id` onto a property.
> 
> ```
> 
> class Lies:
> 
>     def __init__(self):
>         self._agentops_agent_id = uuid4()
>       
> 
>     def __getattr__(self):
>       return "lied"
> 
> 
> 
> Lies()._agentops_agent_id # lied
> ```
> 
> Without a descriptor, there is no way to determine whether the value accessed at `_agentops_agent_id` is genuine, which is exactly what was happening with retrieving the agent_id from call stack: a sea of lies.

Great catch.

teocns:
I can't get `black` to pass, do we have a pre-commit hook to run? might aswell be `black` version/rules mismatches: how are we standardizing that across development environments?

dot-agi:
@teocns @areibman is this good to merge? Will be helpful for #500.

areibman:
Not good to merge. This is failing our integration with CrewAI:
```
(env) (base) ➜  job-posting git:(main) ✗ python main.py
🖇 AgentOps: Session Replay: https://app.agentops.ai/drilldown?session_id=fd72ba4e-8d25-43e4-97b9-dd0fb88baf91
What is the company description?
job
What is the company domain?
jobby
What are the hiring needs?
jobbb
What are specific_benefits you offer?
good job
🖇 AgentOps: Failed to track an agent with the @track_agent decorator.
🖇 AgentOps: Failed to track an agent with the @track_agent decorator.
🖇 AgentOps: Failed to track an agent with the @track_agent decorator.
```

Here's a good example repo to use. Use the job posting example, I find it easiest
https://github.com/AgentOps-AI/crewAI-examples

dot-agi:
@teocns I think you should update the examples with the `track_agent` decorator to demonstrate its usage and that would also help test it out for those agents frameworks.

braelyn-ai:
> I can't get `black` to pass, do we have a pre-commit hook to run? might aswell be `black` version/rules mismatches: how are we standardizing that across development environments?

black is super opinionated and has no configuration. you can run it locally to see what its upset about

teocns:
## Key Changes
1. Improved property name resolution:
   - Added fallback mechanisms when `__set_name__` isn't called
   - Better handling of dynamically added properties
   - Proper handling of `agentops_` prefix stripping

2. Enhanced error handling:
   - Graceful handling of property name resolution failures
   - Better exception handling in stack inspection
   - Clear error messages when property names can't be determined

3. Added comprehensive test suite:
   - Tests for basic property get/set operations
   - Tests for stack frame inspection
   - Tests for inheritance scenarios
   - Tests for error cases and edge conditions

## Technical Details
- The descriptor now properly handles both static class attribute definition and dynamic property addition
- Stack frame inspection is more resilient to exceptions
- Property naming is more consistent with automatic prefix handling
- Added proper type hints and documentation

## Testing
Added comprehensive test suite covering:
- Direct property access
- Nested function calls
- Multiple agents in different stack frames
- Inheritance scenarios
- Error conditions
- Dynamic property addition

teocns:
Big Win. It now comes with general pydantic support 🥳 

For CrewAI, it requires the [following patch](https://gist.github.com/teocns/517435816becc1581250d145bd9eead9); here's why:

For clean code reasons I renamed the property prefix standard to `agentops` (unified), rather than `agent_ops`
examples:
```
agentops_agent_id
_agentops_agent_name
```
Which makes a lot more sense to contain the scope and purpose within a single snake case group `_scope_`

@areibman let's merge this  - let me know if you wanna move back to agent_ops property nomenclature for an express merge

areibman:
Amazing, works for me on several CrewAI and Autogen examples

areibman (approved):
Great stuff

teocns:
@the-praxs  second pair of eyes before merge pls?

dot-agi (approved):
LGTM 🚀

PR #507, opened 2024-11-14
fix: anthropic provider on raw message

teocns:
# 🔍 Review Summary 
 **Purpose**:
   Enhance reliability and response handling logic in the `AnthropicProvider` class within the `agentops/llms/anthropic.py` file.

**Key Changes**:
   - **Refactor**: Reorganized import statements for improved clarity and moved `import json` to the top. Improved response handling logic to correctly parse and log data for `APIResponse` and `LegacyAPIResponse` types. 
   - **Enhancement**: Enhanced logic to handle different response structures, including extracting model and token usage information.
   - **Style**: Made minor formatting changes for better code readability and maintainability.

**Impact**:
   Improves the reliability of the `AnthropicProvider` class by ensuring correct parsing and logging of response data and enhancing the handling of different response structures. 

<details><summary>Original Description</summary>

No existing description found

</details>


teocns:
~Blocked by #510~

Only static analysis checks are failing (black); free to merge afterwards

---

Edit 17/11
Free to merge

agentops/llms/anthropic.py
```diff
@@ -168,6 +196,9 @@ def create_patched_function(is_beta=False):
             def patched_function(*args, **kwargs):
                 init_timestamp = get_ISO_time()
                 session = kwargs.get("session", None)
+                # if is_beta:
+                #     breakpoint()
+
```
areibman: Still needed?

areibman (commented):
Need to ask @the-praxs to double check this but looks good

agentops/llms/anthropic.py
dot-agi: Nope

agentops/llms/anthropic.py
```diff
@@ -118,17 +122,41 @@ async def async_generator():
...
+                llm_event.model = response_data["model"]
+                llm_event.completion = {
+                    "role": response_data.get("role"),
+                    "content": response_data.get("content")[0].get("text") if response_data.get("content") else "",
+                }
+                if usage := response_data.get("usage"):
+                    llm_event.prompt_tokens = usage.get("input_tokens")
+                    llm_event.completion_tokens = usage.get("output_tokens")
+
+            # Han
```
dot-agi: Incomplete comment 👀

agentops/llms/anthropic.py
```diff
@@ -118,17 +122,41 @@ async def async_generator():
...
+            else:
```
dot-agi: Better use an elif and walrus operator for readability. This will help if something changes in the sdk in future.

tox.ini
```diff
@@ -20,6 +20,7 @@ deps =
     types-requests
     psutil
     openai
+    anthropic[bedrock,vertex]
```
dot-agi: Bedrock and Vertex clients are unsupported.

dot-agi (changes requested):
Minor changes needed.

I would remove requirement for the Bedrock and Vertex clients. But we can add support for them in a different PR.

tox.ini
teocns: I couldn't get it to run (locally) with just `anthropic` as dependency, IDK why. I declared bedrock & vertex coz I snatched that dependency configuration from anthropic vision demo own dependencies, which came as `anthropic[bedrock,vertex]`, and it worked....

Does it work for you without the extras?

agentops/llms/anthropic.py
teocns: This if condition is something that I studied a bit in the process.

`if isinstance(response, (APIResponse, LegacyAPIResponse))` is actually redundant here, a simple `not hasattr(response, "model_dump")` would be enough to jump in the block, but I needed to basically emphasize _"rather than relying on `model_dump` - which is too ambiguous - we should consider that we might recieve types `X` and `Y` and probably handle those"_. It's a back-wards compatible condition and it's on purpose structured the way it is

tox.ini
teocns: Anyway I think I added that dep because I was going to write tests which I never did in the end... We can remove the tox dep I think

dot-agi: It won't work since the client and types are different for these. So better to remove this from the dependencies.

dot-agi (approved):
image

Issue #460, opened 2024-10-21
[Docs]: Ollama documentation


areibman:
### 📘 Current State of Documentation

We currently have no sections on how to use Ollama despite the working integration

### 📖 Suggested Improvement

- Docs
- Readme section
- Jupyter notebook

### 🤝 Contribution

- [ ] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the Agentops team to handle this update.

PR #515, opened 2024-11-18
Added simple Ollama example

akhilbandhu:
## 📥 Pull Request

**📘 Description**
Added an ollama example to satisfy Issue https://github.com/AgentOps-AI/agentops/issues/460

**🧪 Testing**
Tested out changes in local UI


examples/ollama_examples/ollama_examples.ipynb
areibman, line 1: image

Either only run `agentops.init()` without `agentops.start_session()` or run `agentops.init(auto_start_session=False)` plus `agentops.start_session()`

areibman, line 1: image
2 things:
1. Restart the kernel and run from top to bottom to ensure it works. Here, the first cell is `[14]` which may indicate the notebook state is wrong
2. Can you clear the outputs from the install cell before committing the save? Makes the notebook easier to read

areibman (changes requested):
Very minor changes, but overall looks great!

docs/v1/examples/examples.mdx
areibman, line 1: Can you create a custom examples page? You can look at the LangChain example as a start.

You probably wouldn't know this, but we have a tool that dynamically injects the outputs of the example notebook into the page. i.e.
image

areibman (changes requested):

docs/v1/examples/examples.mdx
akhilbandhu: @areibman Sure I can do that, what is the name of the tool you mentioned? Didn't want to go around replicating work. Let me know what it is, and I can get that pushed up in the next commit

areibman: Including the line `{/*    SOURCE_FILE: ...` should do the trick. You can copy the LangChain example .mdx file and use that as the starting point

akhilbandhu: Cool beans, I am assuming that that will only run when this is merged with main? Looks like it's a Github action

akhilbandhu: Actually was able to figure out the nbconvert tool and got it working! @areibman Let me know what you think

areibman (approved):
Awesome!!

PR #522, opened 2024-11-21
Independent `get_analytics` method

dot-agi:
## 📥 Pull Request
References #437 by using the latest commits.

**📘 Description**
* Adds a separate `get_analytics` method to enable its usage independently of the session lifecycle
* Adds helper methods for the `get_analytics` method - `_format_duration` and `_format_token_cost` for formatting needs and `_get_response` for fetching the response from the server.
* Adds `token_cost` and `session_url` as attributes to the `Session` class

**🧪 Testing**
* Tested for both single and multiple session(s) with the correct formatting and calculations
* Ensured that no additional requests are made using the `get_response` method


agentops/session.py
```diff
@@ -388,5 +360,72 @@ def wrapper(*args, **kwargs):
...
+    def _format_token_cost(token_cost_d):
+        return (
+            "{:.2f}".format(token_cost_d)
+            if token_cost_d == 0
+            else "{:.6f}".format(
+                token_cost_d.quantize(Decimal("0.000001"), rounding=ROUND_HALF_UP)
+            )
+        )
+
+    def get_analytics(self, response: Response) -> dict[str, Union[Decimal, str]]:
```
areibman: Can you also update the docs to show how to use this method? Will be a helpful way to document the change

areibman (changes requested):

agentops/session.py
dot-agi: Sure!

dot-agi:
When I was writing the documentation, I figured that the current implementation was not effective, since it required a `Response` object from the user to get the analytics. 

Since it is one of the attributes that a user doesn't need to know about, I refactored the code. The code is more readable and modular, and allows the user an easy way to fetch the analytics.

agentops/session.py
dot-agi: I am working on this now, I made a little refactoring.

dot-agi: Done.

areibman (approved):
Awesome work!

PR #525, opened 2024-11-21
Drop support for `python3.8`

### LINKED ISSUE #524: deps: `uv sync --all-extras` fails due to langchain & `python 3.7`
#### Command Used:
```
uv sync --all-extras
```

#### Error Encountered:
```
Using CPython 3.12.2
Creating virtual environment at: .venv
  × No solution found when resolving dependencies:
  ╰─▶ Because the requested Python version (>=3.7) does not satisfy Python>=3.8 and langchain==0.2.14 depends on Python>=3.8.1,<4.0, we can conclude that langchain==0.2.14 cannot
      be used.
      And because agentops[langchain] depends on langchain==0.2.14 and your project requires agentops[langchain], we can conclude that your projects's requirements are unsatisfiable.

      hint: The `requires-python` value (>=3.7) includes Python versions that are not supported by your dependencies (e.g., langchain==0.2.14 only supports >=3.8.1, <4.0). Consider
      using a more restrictive `requires-python` value (like >=3.8.1, <4.0).
```

### Translation

- `langchain==0.2.14` requires `Python>=3.8`
- Agentops `requires-python >= 3.7` (see [pyproject.toml](https://github.com/AgentOps-AI/agentops/blob/8d58b7f3616bbc529ea6fd704882b3fc78a2578b/pyproject.toml#L16))

### Proposal

To `require-python >= 3.9` which is actually (these times) the "oldest" supported version across most packages

teocns:
Fixes #524

Drops support for `py3.8` and lower versions

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -25,7 +25,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.7,3.8,3.9,3.10,3.11,3.12]
+        python-version: [3.9,3.10,3.11,3.12]
 
     steps:
       - uses: actions/checkout@v2
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -114,7 +114,7 @@ We use the following testing packages:
 
 We use tox to automate and standardize testing. Tox:
 - Creates isolated virtual environments for testing
-- Tests against multiple Python versions (3.7-3.12)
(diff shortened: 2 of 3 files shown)


teocns:
Closing in favor of https://github.com/AgentOps-AI/agentops/pull/526 - we may not be ready to drop support for python 3.8 at least until MacOS widely ships with a higher version out of the box

PR #556, opened 2024-12-07
Loosen Dependency Restriction for packaging Version

Wendong-Fan:
## 📥 Pull Request

**📘 Description**
I’ve encountered an issue where the fixed dependency on packaging version 23.2 in AgentOps (version 0.3.12) causes conflicts with other packages that require a different version of packaging.

Would it be possible to update the dependency to allow a range of versions for packaging instead of fixing it to 23.2? For example, using packaging>=23.0 would provide greater flexibility while ensuring compatibility.

close https://github.com/AgentOps-AI/agentops/issues/555

**🧪 Testing**
_Describe the tests you performed to validate your changes._

PR #561, opened 2024-12-09
deps: remove `packaging`; unpinned & ranged versioning

### LINKED ISSUE #545: [Bug]: Stop pinning dependency versions
### Contact Details

_No response_

### 📦 Package Version

0.3.18

### 🎞️ Framework Version

_No response_

### 🔎 Describe the Bug

AgentOps dependencies are largely pinned to a specific version (`==`). This is causing issues with projects that depend on AgentOps and use slightly different versions. When possible, use a range (`>=, <=, ^`)

### 🤝 Contribution

- [X] I'd prefer the AgentOps team to handle this update.

### LINKED ISSUE #555: [Feature]: Loosen Dependency Restriction for `packaging` Version
### 💡 Feature Description and Proposed Solution

I’ve encountered an issue where the fixed dependency on packaging version 23.2 in AgentOps (version 0.3.12) causes conflicts with other packages that require a different version of packaging.

Would it be possible to update the dependency to allow a range of versions for packaging instead of fixing it to 23.2? For example, using packaging>=23.0 would provide greater flexibility while ensuring compatibility.

### 🤔 Related Problem

_No response_

### 🤝 Contribution

- [X] Yes, I'd be happy to submit a pull request with these changes.

teocns:
Closes #545
Closes #556
Fixes #555

---

- `packaging` does not have to be verbosely specified because:
  1. It is already an implicity dependency upon `setuptools`
  2. Even if the above wouldn't be true, it still would not belong under runtime `[dependencies]`. 
  For further info, https://packaging.python.org/

- Restored loosen `psutil` loosen dep from #491

- Capped all dependencies to a maximum ver. pointing to the latest stable (there are a number of reasons to do that, among security and also performance)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,9 +21,8 @@ classifiers = [
 ]
 dependencies = [
     "requests>=2.0.0,<3.0.0",
-    "psutil==5.9.8",
-    "packaging==23.2",
-    "termcolor>=2.3.0", # 2.x.x tolerant
+    "psutil>=5.9.8,<6.1.0",
+    "termcolor>=2.3.0,<2.5.0",
     "PyYAML>=5.3,<7.0",
     "opentelemetry-api>=1.22.0,<2.0.0", # API for interfaces
     "opentelemetry-sdk>=1.22.0,<2.0.0", # SDK for implementation


dot-agi (approved):
LGTM

PR #585, opened 2024-12-16
[Fix] Updated notebooks that publish to examples docs pages

dot-agi:
# Updated notebooks that publish to examples docs pages

# Purpose
* To make the example notebooks and docs pages clean and consistent for a good developer UX.
* To sync the example notebook and the respective example docs pages for various integrations.
* Clean the example notebooks

# Changes
| Example Docs | Jupyter Notebook |
|--------|--------|
| `v1/examples/langchain` | langchain_examples |
| `v1/examples/multion` | Autonomous_web_browsing |
| `v1/examples/simple_agent` | openai-gpt |
| `v1/examples/camel` | camelai-multi-agent-example |

## Miscellaneous
* Modified Anthropic example notebooks and corresponding README
* Merged CAMEL-AI notebook examples into a single notebook `camel-simple-examples.ipynb`
* Added a new notebook for the langchain v0 example `langchain_v0_example.ipynb`
* Modified agentops key variable name in xAI example notebook


dot-agi:
@TKTSWalker have a look here as I said before.

TKTSWalker:
Hey, just giving things a look right now! There seems to be be 2 blocks for CamelAI on the intiial readme to begin

image

image

Mine's is the second one although with a broken link (from the links being set to the wrong place, examples/camel is now examples/camel_examples)! I mainly formatted it that way to match the formatting of the rest for readability (well with the lower agents anyways), the Installation part is definitely better than what I placed for sure though!

I'm also surprised there isn't some sort of general glossary for example files if you guys would rather not have a reference  to the github side journals! Adding an area to the bottom can help greatly with this, having names for each integration and a quickview of what's within!

Then on the journal end things look mostly fine (although I have been awake for the last 48h working on a few projects so I might be missing details here)! One thing I always like to do when making journals though is simplify terms and logic to best fit the example (although I should use the technical terms a little more)! So I was thinking maybe I could spin my hand and make examples with the simple terms for newcomers and once again for pros! As an example, one snippet you put was;

"first we'll create an instance of the AO LangchainCallbackHandler. After the handler is initialized, a session will be recorded automatically."

This could be mirrored with something like

"We will create an object for holding data called "Variable Name". When we call it, AgentOps will automatically record for us!"

Or something like that, just throwing ideas around there more than anything else though! 

The CamelAI page has more broken references in general relating to files, although I like how you implemented the training data link way more!

I'll be a *little* busy today as i've gotta work on something for a little project of mines and handle a few other things, but if you want me to hop on fixing these through a push request to your repo it's something I can do!

dot-agi:
Thanks for the review!

For the main `README.md`, I am going to ask the team what action to take since it's not updated and will be bloated if we put too much info there. Everything is on the docs so one can refer that for any information.

In case of the examples/journals, I try to create a balance between technical and simplicity, since I see it from the eyes of a noob. Also, I tend to keep it professional and straight-forward. My minimum expectation is that the user must have an understanding of the basic programming concepts and can read the technical dialect we speak in 👀 

The broken references are due to being pointed to `main` so it will work after the merge.

areibman (approved):

TKTSWalker:
No problem! I was thinking it could be nice if the main readme was more visible! I could hop on Figma and make a few banners; then when you want more information on a specific you could use the dropdown! Might it be a good idea to even just have a general description under each and have a minimized readme as a sort of "General" as well as the other readme for each? It'd make going through things much easier and I know there are tools to create a sort of wiki within a Github repo itself!

On the examples/journals side, that makes complete sense! I was thinking more like it might be a little hard to remember certain terms when you're starting out but that should realistically not be a huge barrier given they're also self explanatory more often than not!

The link thing makes sense!

Would you like me to draft up a readme page with the thoughts I have in mind?

PR #601, opened 2024-12-23
 Fix test_host_env to handle different psutil versions

### LINKED ISSUE #600: Test errors in `tests/test_host_env.py` caused by psutil
This happens due to `psutil` having a soft requirement and apparently there's a breaking change somewhere in between
```
tests/test_host_env.py:26: in test_disk_info
    self.assert_disk_info()
tests/test_host_env.py:34: in assert_disk_info
    disk_info = host_env.get_disk_details()
agentops/host_env.py:113: in get_disk_details
    partitions = psutil.disk_partitions()
tests/test_host_env.py:23: in <lambda>
    @patch("psutil.disk_partitions", new=lambda: [mock_partitions()[0]])
tests/test_host_env.py:10: in mock_partitions
    sdiskpart(device="/dev/sda1", mountpoint="/", fstype="ext4", opts="rw,relatime"),
E   TypeError: sdiskpart.__new__() missing 2 required positional arguments: 'maxfile' and 'maxpath'
```

The is failing locally but not in CI because of different versions of `psutil`. In newer versions (5.9.0+), `sdiskpart` requires additional fields (`maxfile` and `maxpath`) that weren't needed in older versions.

teocns:
Fixes #600 

## Problem
The test was failing locally but passing in CI due to differences in psutil versions. Newer versions of psutil (5.9.0+) require additional fields (`maxfile` and `maxpath`) in the `sdiskpart` namedtuple that weren't present in older versions.

## Solution
Implemented a version-agnostic approach using try/except to handle both new and old psutil versions:
1. First attempts to create `sdiskpart` with new fields (maxfile, maxpath)
2. Falls back to old format if TypeError is raised
3. Removed complex dynamic field detection in favor of simpler fallback pattern

## Changes
- Simplified `mock_partitions()` to use try/except pattern
- Removed unnecessary namedtuple dynamic field detection
- Added type ignore comments for linter compatibility
- Maintained test functionality across different psutil versions

### DIFF
--- a/tests/test_host_env.py
+++ b/tests/test_host_env.py
@@ -1,15 +1,27 @@
 from unittest.mock import patch
 from agentops import host_env
+import psutil
 
 # noinspection PyProtectedMember
 from psutil._common import sdiskpart, sdiskusage
 
 
 def mock_partitions():
-    return [
-        sdiskpart(device="/dev/sda1", mountpoint="/", fstype="ext4", opts="rw,relatime"),
-        sdiskpart(device="z:\\", mountpoint="z:\\", fstype="ntfs", opts="rw,relatime"),
-    ]
+    # Try to create with new fields first, fall back to old format if it fails
+    try:
+        return [
+            sdiskpart(  # noqa: E501
+                device="/dev/sda1",
+                mountpoint="/",
+                fstype="ext4",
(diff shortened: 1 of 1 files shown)


dot-agi (approved):
Approved 🤌🏼

PR #606, opened 2024-12-27
Drop support for py38 & uv chore upgrade

teocns:
## 🎯 Purpose
Replace tox with uv for testing and dependency management to improve CI speed, simplify configuration, and modernize our tooling. Drop support for Python 3.8 (and lower) as it reached its [EOL](https://devguide.python.org/versions/)

image

## 🤔 Reasoning

1. **Why UV?**
   - 5-10x faster dependency resolution than pip
   - Better caching support in CI
   - Modern Python packaging features
   - Simpler configuration in pyproject.toml
   - Active development and community support

2. CI Testing:
- Workflow will automatically test against Python 3.9-3.13
- Coverage reports will be generated and uploaded to Codecov

## 📈 Performance Impact
- CI execution time expected to improve by 40-60%
- Dependency resolution time reduced by 5-10x
- Improved caching reduces subsequent run times

## 🔍 Additional Notes
- UV is maintained by the Ruff team, ensuring long-term support
- Migration is backward compatible - users can still use pip if preferred
- No changes to test code required


teocns:
- Reverted some devin tests that are poorly implemented and misusing the async loop causing CI to timeout

- Reverted https://github.com/AgentOps-AI/agentops/commit/82f9bb989e726c88e84265fc2399093cdcdfecf5 as support for <= 3.8 is dropped and it defeats the purpose of the commit

dot-agi (approved):
LGTM 🚀

Issue #607, opened 2024-12-27
[Bug]: Multiple return values from a decorated function become a list


zli11010:
### 📦 Package Version

0.3.18

### 🔎 Describe the Bug

If I do something like
```python
@agentops.record_action("Action")
def some_action():
    return 0, 1
```
then the output becomes the list `[0, 1]`.

This is an issue because I'm using a repository where they check whether the output from a function is a specific subclass of `tuple`, so when I added the decorator to the function the code broke.

### 🤝 Contribution

- [X] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the AgentOps team to handle this update.

zli11010:
I fixed the bug in this pr: https://github.com/AgentOps-AI/agentops/pull/608

areibman:
Taking a look at the PR, thanks! @zli11010 are you using a specific project or rolling out an agent of your own?

PR #608, opened 2024-12-27
[Fix] Removed unnecessarily changing tuple output to list output

zli11010:
## 📥 Pull Request

**📘 Description**
Fixed this bug: https://github.com/AgentOps-AI/agentops/issues/607

**🧪 Testing**
Ran my code that uses AgentOps and now it works. (Changing list to tuple doesn't affect what gets logged.)


agentops/decorators.py
```diff
@@ -67,10 +67,6 @@ async def async_wrapper(*args, session: Optional[Session] = None, **kwargs):
                 try:
                     returns = await func(*args, **kwargs)
 
-                    # If the function returns multiple values, record them all in the same event
-                    if isinstance(returns, tuple):
-                        returns = list(returns)
-
```
areibman, line 73: Instead of removing these functions, we can instead do what we intended at fist-- just return a tuple

```
if istinatnce(returns, tuple):
    returns = tuple(returns)
```

Removing these lines of code altogether will have unintended consequences.

areibman (dismissed):

areibman:
Left a review @zli11010-- should be an easy fix! Please let me know if it still causes problems with the repo you're trying

teocns:
@areibman can we have more documentation around the expected functionality of `returns`? I too had a hard time figuring out, maybe some docstrings will do 👍

agentops/decorators.py
zli11010: May I ask what the suggested change does? It seems (in most cases) equivalent to removing the two lines of code.

Unfortunately your suggested change would still break the code I'm using, since the code checks the output of the function is an instance of a *specific subclass* of `tuple` (more specifically, a custom `namedtuple`). Maybe one could argue that this is bad practice, but I believe it's more important that a decorator that does logging doesn't change the output of the function.

zli11010:
@areibman Since you had concerns about deleting the lines, I tried to keep the original semantics with the only change being fixing the fact that the type casting modified the output of the original function. Hopefully that's better!

agentops/decorators.py
teocns: > but AFAIK the only reason the tuple was being cast to a list was so that it can be dumped into a JSON, but json.dumps actually handles that automatically.

Indeed that should be handled by a `JSONEncoder` which is where these sorts of casts belong

dot-agi:
What's the status of this PR?

areibman:
@teocns can you take a stab at completing this? Just want to make sure we don’t have any downstream effects if we merge

teocns:
The removal of the tuple-to-list conversion seems to be the right approach; the conversion was actually unnecessary since:

- Our serialization layer (`safe_serialize` and `filter_unjsonable`) already handles tuples correctly
- The conversion was happening too early (at decorator level) rather than at serialization time

Added some tests to ensure `@record_*` decorators behave correctly in regards to the issue, especially proving the json serialization utilities already handle tuple types correctly without needing to convert them to lists first, proving the point that the conversion was unnecessary :) 
   
   
If this gets merged, we should open a ticket to unify:
- [`test_record_action.py`](https://github.com/AgentOps-AI/agentops/blob/80d4bf59cc7b0faef43b8e251b2002a3ee40f65f/tests/test_record_action.py)
- [`test_record_tool.py`](https://github.com/AgentOps-AI/agentops/blob/80d4bf59cc7b0faef43b8e251b2002a3ee40f65f/tests/test_record_tool.py)

into:
- [`test_decorators.py`](https://github.com/AgentOps-AI/agentops/blob/64fc5ab7c4a8e3e97394a195e1394b5c381a17a3/tests/test_decorators.py)

--- 

> Rationale: Both `record_action` and `record_tool` are decorators with similar functionality and testing patterns.
Having their tests in separate files creates unnecessary fragmentation.

PR #666, opened 2025-01-31
fixes: prevented block on start_session and cold_session

Abraham599:
## 📥 Pull Request

**📘 Description**
these are the steps that I took:-

i)(fix for start_session) JWT fetch will happen in a background thread and queue events until the JWT is available 
ii) (fix for end_session) handle the flush in a background thread


areibman:
@teocns After the logs fix can you take a look at this? 

Thanks for raising the pr @Abraham599 !!

PR #667, opened 2025-02-03
integrated langgraph callback handler along with a test file

Abraham599:
## 📥 Pull Request

**📘 Description**
This PR integrates langgraph callback handler and also includes a test file. This implementation follows the same approach used for langchain


dot-agi:
Thanks for the PR! 

Can you also add example notebook(s) and the associated docs?

PR #668, opened 2025-02-04
Logcap

teocns:
Capture terminal output to be displayed in the agentops dashboard. Session-based capture functionality is currently under development. Captures will be displayed in the AgentOps dashboard under "Terminal".

At the moment we capture:

- logging, i.e `logging.info()`, `logging.error()`, etc.
- stdout/stderr, i.e `print()`

---

Key changes in this PR:

1. **HTTP Client Refactoring**
- Consolidated HTTP methods (GET, POST, PUT, DELETE) into a single [`_make_request`](https://github.com/AgentOps-AI/agentops/blob/logcap/agentops/http_client.py#L115) method
- Added PUT and DELETE methods in [HttpClient](https://github.com/AgentOps-AI/agentops/blob/logcap/agentops/http_client.py#L182-L201)
- Made payload parameter optional

2. **New Logging & Instrumentation Features**
- Added [instrumentation.py](https://github.com/AgentOps-AI/agentops/blob/logcap/agentops/instrumentation.py) for OpenTelemetry setup and configuration
- Added [log_capture.py](https://github.com/AgentOps-AI/agentops/blob/logcap/agentops/log_capture.py) with classes for capturing stdout/stderr and logging
- Added session-specific logging with ANSI color support
- Added log export functionality to send logs to AgentOps backend

3. **Session Updates**
- Added logging components to Session class
- Added log exporter for sending logs to backend
- Updated session cleanup to handle logging components

4. **Testing**
- Added [tests/unit/test_log_capture.py](https://github.com/AgentOps-AI/agentops/blob/logcap/tests/unit/test_log_capture.py)
- Updated [conftest.py](https://github.com/AgentOps-AI/agentops/blob/logcap/conftest.py) with logging-related mocks. Introduced `agentops_session` and `agentops_init` which promotes auto setup/teardown.


teocns:
The current implementation has a potential issue: it doesn't guarantee that logs from one session won't leak into another when running concurrently.

This needs to be addressed with callstack lookup or other methods. @areibman: otherwise, how does agentops actually know which session to assign llm events to? would the same mechanism work?

PR #713, opened 2025-02-28
Redesign pytest

### LINKED ISSUE #706: Create integration tests for v0.4 providers
**IMPORTANT:** Open a PR basing your branch off `redesign` to merge onto `redesign`

Dwij1704:
## 📥 Pull Request

**📘 Description**
Added instrumentation tests for the Anthropic, OpenAI  and Haystack. These tests ensure that API calls are tracked and verified using spans.

**🧪 Testing**
To validate these changes, run any integration test. The tests utilize VCR to record API calls and cassettes for replaying them. Snapshots are used to verify the expected outputs.


tests/integration/test_anthropic_instrumentation.py
```diff
@@ -0,0 +1,41 @@
...
+                "attributes": dict(span.attributes),
+                "status": span.status.is_ok
+            }
+            for span in finished_spans
+        ]
+        
+        spans_json = json.dumps(spans_data, indent=2)
+        
+        # Verify the spans using snapshot
+        snapshot.assert_match(spans_json, 'anthropic_spans.json')
```
teocns: this is not looking right, is it? 

snapshots are supposed to _inline_ override the assertion expression, generating the body of the span after the second run 

1.Before running snapshotter:
assert data = snapshot()

2.After running snapshotter:

assert data = ["snapshotter edited the test file and generated this data"]

teocns (changes requested):

tests/integration/test_anthropic_instrumentation.py
Dwij1704: Initially we would need to create a snapshot manually using --snapshot-update flag in pytest, also it only accepts str or bytes that explains the json format
image

Dwij1704: Also the 'Snapshot' object is not callable

tests/integration/test_anthropic_instrumentation.py
```diff
@@ -0,0 +1,88 @@
+import pytest
+import os
+os.environ["OPENAI_API_KEY"] = "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+os.environ["ANTHROPIC_API_KEY"] = "000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"
+os.environ["AGENTOPS_API_KEY"] = "00000000-0000-0000-0000-000000000000"
```
teocns: use fixtures for these and exclude them from snapshots

tests/integration/test_anthropic_instrumentation.py
```diff
@@ -0,0 +1,88 @@
...
+import vcr
+import json
+import anthropic
+from inline_snapshot import snapshot
+
+@vcr.use_cassette('tests/integration/cassettes/test_anthropic_instrumentation.yaml')
+def test_anthropic_instrumentation(exporter):
+    """Test that Anthropic API calls are tracked in spans using inline snapshots"""
+
+    session = agentops.start_session()
```
teocns: instead of starting the session manually, use fixture `agentops_session` which already handles session start / end, and requires no further action

teocns (changes requested):

PR #750, opened 2025-03-08
Get tests passing

tcdent:
Tests were not able to access `AGENTOPS_API_KEY` because mock env clears all environment variables. Add a random uuid to the config mock.

### DIFF
--- a/tests/fixtures/config.py
+++ b/tests/fixtures/config.py
@@ -7,7 +7,10 @@ from pytest_mock import MockerFixture
 
 @pytest.fixture(autouse=True)
 def mock_env():
-    with mock.patch.dict(os.environ,clear=True) as mock_env:
+    """Clear environment but preserve AGENTOPS_API_KEY if it exists."""
+    api_key = os.environ.get("AGENTOPS_API_KEY")
+    with mock.patch.dict(os.environ, clear=True) as mock_env:
+        os.environ["AGENTOPS_API_KEY"] = api_key
         yield mock_env
 
 
--- a/tests/fixtures/config.py
+++ b/tests/fixtures/config.py
@@ -1,4 +1,5 @@
 import os
+import uuid
 from unittest import mock
 
 import pytest
@@ -8,9 +9,8 @@ from pytest_mock import MockerFixture
 @pytest.fixture(autouse=True)
 def mock_env():
     """Clear environment but preserve AGENTOPS_API_KEY if it exists."""
-    api_key = os.environ.get("AGENTOPS_API_KEY")
     with mock.patch.dict(os.environ, clear=True) as mock_env:
-        os.environ["AGENTOPS_API_KEY"] = api_key
+        mock_env["AGENTOPS_API_KEY"] = uuid.uuid4().hex
         yield mock_env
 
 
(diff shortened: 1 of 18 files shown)


braelyn-ai:
all hail uv

braelyn-ai (approved):

tcdent:
I added a couple more fixes to this while I was getting tests to pass:

- [x] Suppress the barrage of otel log messages at the end of tests. 
- [x] Get the `agentops.config` tests to pass.

teocns:
Dependency-group is a valid uv key. When you run 'uv sync',pyproject is configured to pick up onto both dev,test groups automatically

teocns:
When working with uv the intended developemt usage is 'uv run' for everything. Uv run pytest, uv run file.py. This automatically sets up the uv ecosystem and .venv for you

teocns:
`/tests/unit` tests are not meant to make real outbound requests whatsoever. See `mock_req` fixture

for real outbound requests, place your tests either under `/tests/integration/` for a handy pre-configured VCR environment, or `/tests/smoke` for absolute real environment tests

tcdent:
@teocns Reverted changes to pyproject.toml, uses dummy api key.

tests/fixtures/config.py
```diff
@@ -7,7 +8,9 @@
 
 @pytest.fixture(autouse=True)
 def mock_env():
-    with mock.patch.dict(os.environ,clear=True) as mock_env:
+    """Clear environment but preserve AGENTOPS_API_KEY if it exists."""
+    with mock.patch.dict(os.environ, clear=True) as mock_env:
+        mock_env["AGENTOPS_API_KEY"] = uuid.uuid4().hex
```
teocns, line 13: let's encapsulate `api_key` into a fixture (this is already done in `tests/unit/conftest.py`). 

That way, the api key being used throughout tests can be referenced anywhere in tests as:



```
@pytest.fixture # in tests/unit/conftest.py
def api_key():
  return 'test-api-key'


def test_anything(api_key)
  api_key == 'test-api-key'

```

tests/unit/test_agentops_init.py
```diff
@@ -174,7 +174,7 @@ def test_kwargs_override_env_vars(agentops_config, mock_config):
 def test_no_api_key_raises_exception():
     """Test that an exception is raised when no API key is provided"""
     with pytest.raises(NoApiKeyException):
-        agentops.init()
+        agentops.init(api_key=False)  # have to use `False` because `None` gets filtered
```
teocns, line 177: awesome trick!

teocns (changes requested):

tests/fixtures/config.py
teocns: FYI better use literal `test-api-key` since we can easily identify it within tests. If it were to be an actual `uuid4()`, we couldn't intuitively distinguish it, hence can't determine if the mocks are working properly, or an `api_key`'s comes from the real environment

tcdent: Tried returning it as 'test-api-key' from the existing test fixture, but it doesn't get assigned before most of the tests run. 

Also tried returning it from `mock_env` as a string (`"test-api-key"`) but the tests need a UUID to pass. 

This is still one of the only ways I've been able to get this to pass.

teocns: interesting.

Yes this is a bit tricky, let me take a look

teocns:
much inefficiency dun care

PR #762, opened 2025-03-09
Redesign session registry mixin

teocns:
- Moved session registry functions from `session.py` to `registry.py`
- Created `SessionRegistryMixin` to encapsulate registry-related functionality
- Delegated registry calls to the mixin: `Session` calls `super().start()` and `super().end()`, which mixin handles.
- implement proper session registry tests
- Updated existing `test_session.py` to use the new module paths


Ashley6T (commented):
Txt

PR #792, opened 2025-03-11
Integration test for dev branch. 

tcdent:
Creates an isolated test environment for each framework with `uv run` and tests across multiple Python versions and multiple framework package versions. 

Run with `bin/tests/test-matrix-integration` for the full suite. Other flags listed inside command file. 

 - [ ] Had to disable new integration tests. TODO re-enable @teocns tests.


teocns:
please change the base to `dev` instead of `main`

PR #843, opened 2025-03-14
fix: show session url when using `agentops.init` only

### LINKED ISSUE #838: URL doesn't print in terminal when running with only `agentops.init`

dot-agi:
Closes #838


teocns:
does it not work on `dev` branch?

Nevertheless this does not go in the right direction

dot-agi:
Does not work on `dev`.

PR #863, opened 2025-03-17
Updated unit tests for decorators by adding workflow and task nesting validation.

Dwij1704:
## 📥 Pull Request

**📘 Description**
Introduced new test case to verify proper span creation and hierarchy for workflows and tasks, ensuring accurate tracing and span relationships.

**🧪 Testing**
All test case passed

image

### DIFF
--- a/tests/unit/sdk/test_decorators.py
+++ b/tests/unit/sdk/test_decorators.py
@@ -8,6 +8,7 @@ from opentelemetry.sdk.trace import ReadableSpan
 from agentops.sdk.decorators import agent, operation, session, workflow, task
 from agentops.semconv import SpanKind
 from agentops.semconv.span_attributes import SpanAttributes
+from agentops.semconv import SpanAttributes
 from tests.unit.sdk.instrumentation_tester import InstrumentationTester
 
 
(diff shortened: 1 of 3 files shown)


tests/unit/sdk/test_decorators.py
```diff
@@ -478,4 +478,93 @@ def test_complex_session():
...
+        workflow_span = None
+        process_task = None
+        transform_task = None
+        
+        for span in spans:
+            if span.attributes and span.attributes.get('agentops.operation.name') == 'data_processing_workflow':
+                workflow_span = span
+            elif span.attributes and span.attributes.get('agentops.operation.name') == 'process_input':
+                process_task = span
+            elif span.attributes and span.attributes.get('agentops.operation.name') == 'transform_data':
```
teocns: Let's use semconv if the decorators also use semconv internally

Dwij1704: Modified span attribute keys in decorators to use constants from SpanAttributes class. Updated unit tests for the same

dot-agi (approved):

PR #879, opened 2025-03-24
Log deeplink to trace on AgentOps dashboard.

tcdent:
TODO
- [x] Verify the value returned by `trace_id_to_uuid` is the correct format for the dashboard. *This needs to be a hex string, but not a UUID. *

### DIFF
--- a/agentops/helpers/dashboard.py
+++ b/agentops/helpers/dashboard.py
@@ -2,11 +2,9 @@
 Helpers for interacting with the AgentOps dashboard. 
 """
 from typing import Union
-from uuid import UUID
 from termcolor import colored
 from opentelemetry.sdk.trace import Span, ReadableSpan
 from agentops.logging import logger
-from agentops.sdk.converters import trace_id_to_uuid
 
 
 def get_trace_url(span: Union[Span, ReadableSpan]) -> str:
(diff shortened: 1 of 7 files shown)


agentops/helpers/dashboard.py
```diff
@@ -0,0 +1,41 @@
...
+Helpers for interacting with the AgentOps dashboard. 
+"""
+from typing import Union
+from termcolor import colored
+from opentelemetry.sdk.trace import Span
+from agentops.logging import logger
+from agentops.sdk.converters import trace_id_to_uuid
+
+
+APP_URL = "https://app.agentops.ai"
```
dot-agi: I think this shouldn't be hardcoded as we have a separate URL for the development environment. 

This endpoint URL is read from the `.env` file if none is provided when initializing the client.

dot-agi (changes requested):
One comment regarding the `APP_URL`.

Tested it, URL looks good!

agentops/helpers/dashboard.py
tcdent: Do we have app URL in the config object?

dot-agi: We have [`endpoint`](https://github.com/AgentOps-AI/agentops/blob/c19d8f940bc8deae4e03eadf396d1a8990c9df7e/agentops/config.py#L43):

```python
endpoint: str = field(
    default_factory=lambda: os.getenv("AGENTOPS_API_ENDPOINT", "https://api.agentops.ai"),
    metadata={"description": "Base URL for the AgentOps API"},
)
```

tcdent: Added it to the config.

tcdent:
Updated this to generate hex strings which matches the dashboard format (not UUID).

dot-agi (approved):

PR #885, opened 2025-03-25
Add LLM_CONTENT_COMPLETION_CHUNK attribute to SpanAttributes class

Dwij1704:
## 📥 Pull Request

**📘 Description**
When running chat.completions using `stream=True`, it throws out 
```An error occurred: type object 'SpanAttributes' has no attribute 'LLM_CONTENT_COMPLETION_CHUNK'```

**🧪 Testing**
Added LLM_CONTENT_COMPLETION_CHUNK attribute and tested using the following script
```
import os
from openai import OpenAI
load_dotenv()
import agentops
agentops.init()
agentops.start_session(tags=["openai_chat"])

client = OpenAI(
    api_key=os.getenv("OPENAI_API_KEY")  # Make sure to set your API key in environment variables
)

def chat_with_gpt():
    try:
        # Create a streaming chat completion
        stream = client.chat.completions.create(
            model="gpt-3.5-turbo",
            messages=[
                {"role": "system", "content": "You are a helpful assistant."},
                {"role": "user", "content": "What is the capital of France?"}
            ],
            temperature=0.7,
            max_tokens=100,
            stream=True  # Enable streaming
        )
        
        # Print the response as it streams
        print("Assistant's response: ", end="", flush=True)
        for chunk in stream:
            if chunk.choices[0].delta.content is not None:
                print(chunk.choices[0].delta.content, end="", flush=True)
        print()  # New line at the end
        
    except Exception as e:
        print(f"An error occurred: {e}")

if __name__ == "__main__":
    chat_with_gpt() 
```

### DIFF
--- a/agentops/semconv/langchain.py
+++ b/agentops/semconv/langchain.py
@@ -25,6 +25,9 @@ class LangChainAttributes:
     # Session attributes
     SESSION_TAGS = "langchain.session.tags"
     
+    LLM_NAME = "langchain.llm.name"
+    LLM_MODEL = "langchain.llm.model"
+    
     # Chain attributes - specific to LangChain
     CHAIN_NAME = "langchain.chain.name"
     CHAIN_TYPE = "langchain.chain.type"
--- a/agentops/semconv/span_attributes.py
+++ b/agentops/semconv/span_attributes.py
@@ -40,8 +40,8 @@ class SpanAttributes:
 
     # Content
     LLM_PROMPTS = "gen_ai.prompt"
-    #LLM_COMPLETIONS = "gen_ai.completion"  # DO NOT SET THIS DIRECTLY
-    #LLM_CONTENT_COMPLETION_CHUNK = "gen_ai.completion.chunk"
+    LLM_COMPLETIONS = "gen_ai.completion"  # DO NOT SET THIS DIRECTLY
+    LLM_CONTENT_COMPLETION_CHUNK = "gen_ai.completion.chunk"
 
     # Response attributes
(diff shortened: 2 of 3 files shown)


tcdent (approved):
Good catch!

Dwij1704:
> Good catch!

@tcdent 
We heard back from VRSEN AI regarding Langchain callback handlers, they said its broken due to `type object 'SpanAttributes' has no attribute 'LLM_COMPLETIONS'`

@the-praxs was going through the implementation and he noticed that you have specifically mentioned `NEVER manually set the root completion attributes (`SpanAttributes.LLM_COMPLETIONS` or "gen_ai.completion")` `Let OpenTelemetry backend derive these values from the detailed attributes` contrary also commented out these attributes from semcnv.

Can you give me some clarity on how you wish to manage both of these conventions, manually or via semconv?

dot-agi:
This catch is a good reminder to have tests for attributes in some way. The order of merge caused this unexpected error.

Dwij1704:
I fixed the error, but when testing it, I can't see token usage in any of the spans for Agents SDK
https://jaeger.agentops.cloud/trace/c9a45998d27d289fa35220a37f6fcac1
Langchain is working as expected
@the-praxs @tcdent Can you confirm if it used to set those attributes before or not?

dot-agi:
> I fixed the error, but when testing it, I can't see token usage in any of the spans for Agents SDK https://jaeger.agentops.cloud/trace/c9a45998d27d289fa35220a37f6fcac1 Langchain is working as expected @the-praxs @tcdent Can you confirm if it used to set those attributes before or not?

Token usage fixed in #888

PR #888, opened 2025-03-25
Fix token counts in OpenAI Agents.

tcdent:
Update attribute extraction to support dict as well as object.

Addresses #886


dot-agi:
Unit tests failing - culprit is in the openai tests.

tcdent:
Tests pass now. We blew up the type checker with the LangGraph merge :/

dot-agi (approved):

PR #899, opened 2025-03-27
Ensure auto started sessions behave properly

### LINKED ISSUE #894: agentops.init() Not Creating Session Span by Default; start_session() Creates Sibling Span Without Context Propagation
When calling `agentops.init()`, a session span is not created by default. This forces us to manually call `agentops.start_session()` to generate a session span. However, even when using `agentops.start_session()`, the generated session span is not propagated correctly and appears as a sibling span rather than as a parent of subsequent spans.

**Steps to Reproduce:**

1. Call `agentops.init()` without calling `agentops.start_session()`.  
   - **Observation:** No session span is created (see [[trace](https://jaeger.agentops.cloud/trace/631a65b1dfeec4c30cddcc1e741876c2)](https://jaeger.agentops.cloud/trace/631a65b1dfeec4c30cddcc1e741876c2)).

2. Call `agentops.start_session()` after `agentops.init()`.  
   - **Observation:** A session span is created, but it becomes a sibling span and does not propagate the context to subsequent spans (see [[trace](https://jaeger.agentops.cloud/trace/4eeb8f9c2487a5557a71b1b4f51c29b2)](https://jaeger.agentops.cloud/trace/4eeb8f9c2487a5557a71b1b4f51c29b2)).

**Expected Behavior:**

- A call to `agentops.init()` should automatically create a session span, or at least the manual call to `agentops.start_session()` should correctly propagate the session context so that the session span is the parent for subsequent spans.

### LINKED ISSUE #898: Ensure session auto-start behaves correctly

tcdent:
- [x] Unit tests for session manual & auto start
- [x] Ensure session state is handled properly
- [ ] Verify root spans are being exported correctly

### DIFF
--- a/agentops/__init__.py
+++ b/agentops/__init__.py
@@ -1,11 +1,17 @@
-from typing import Any, Dict, List, Optional, Union
+from typing import List, Optional, Union
+from agentops.client import Client
 
-from agentops.legacy import ActionEvent, ErrorEvent, ToolEvent, start_session, end_session
-
-from .client import Client
 
 # Client global instance; one per process runtime
 _client = Client()
+
+
+def get_client() -> Client:
+    """Get the singleton client instance"""
+    global _client
+
+    return _client
+
 
 def record(event):
     """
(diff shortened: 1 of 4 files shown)


Dwij1704:
@tcdent Are you planning to fix #894 with this PR?

tcdent:
Will address #894 in a separate PR after internal discussion. @the-praxs ready for review.

dot-agi (approved):
I find this ok but I am more concerned about thread safety since opentelemtry context management uses threads to pass spans.

PR #900, opened 2025-03-28
OpenAI Agents voice support

### LINKED ISSUE #884: OpenAI Agents voice support
Add semantic conventions for voice messages. 

Explore options for storing base64 encoded data outside of span attributes.

tcdent:
- [x] Captures span data for transcriptions & speech.
- [x] Uploads audio data to AgentOps backend.

Example agent script at: `examples/agents-example/voice_transcription.py`


agentops/helpers/objects.py
```diff
@@ -0,0 +1,14 @@
...
+if TYPE_CHECKING:
+    from agentops.client.api.types import UploadedObjectResponse
+
+
+def upload_object(body: bytes) -> 'UploadedObjectResponse':
+    """Upload an object to the agentops server."""
+    from agentops import get_client
+    
+    client = get_client()
+    return client.api.v4.upload_object(body)
```
teocns: Why let `upload_object` be in `agentops.helpers` and not under `agentops.client.versions.v4::upload_object`?

teocns (changes requested):

agentops/helpers/config.py
```diff
@@ -0,0 +1,31 @@
...
+def get_tags_from_config() -> List[str]:
+    """Get tags from the global configuration.
+    
+    Returns:
+        List of tags if they exist in the configuration, or empty list
+    """
+    config = get_config()
+    if config.default_tags:
+        return list(config.default_tags)
+    return []  # Return empty list for empty tags set
```
teocns: Same for these. We're adding extra `helpers` complexities when we could just `from agentops.config import get_config` - same way as we do for `from agentops import get_client`

> half-correct example here, because `get_client`'s implementation should be moved under `agentops.client::get_client` and import forwarded in `agentops.__init__` for correctness

teocns (dismissed):

agentops/helpers/config.py
tcdent: These helpers sidestep two fatal design flaws:

The entire application state is stacked onto a single variable: `_client` which is located in `/__init__.py`. 

1. You cannot import this variable without causing a circular import. 
2. It's a private variable (which is why I made `get_client` part of the public API again in #899)

These helpers sidestep the issue and keep function-level imports to a minimum.

teocns: This is a case for isolating `_client` into `agentops.client.client::_client` as well as `agentops.client.client::get_client` which breaks the circular import issue, and also a potentially consistent design approach

examples/agents-example/voice_transcription.py
dot-agi, line 1: There are two example directories for the Agents SDK - `agents-example` and `agents-examples`. They should be merged into a single one.

agentops/client/api/types.py
```diff
@@ -5,10 +5,18 @@
...
 
 
 class AuthTokenResponse(TypedDict):
     """Response from the auth/token endpoint"""
 
     token: str
     project_id: str
+
+
+class UploadedObjectResponse(BaseModel):
```
dot-agi, line 18: Must also have object id, name is optional though highly recommended.

agentops/helpers/config.py
dot-agi: I agree. This has to be unified in the client config and nowhere else.

agentops/client/api/versions/v4.py
```diff
@@ -0,0 +1,71 @@
...
+    def set_auth_token(self, token: str):
+        """
+        Set the authentication token for API requests.
+
+        Args:
+            token: The authentication token to set
+        """
+        self.auth_token = token
+    
+    def prepare_headers(self, custom_headers: Optional[Dict[str, str]] = None) -> Dict[str, str]:
```
dot-agi, line 26: Duplicate method of the same name as `agentops.client.api.base.prepare_headers`.

Use that method and modify the code aptly for `V3Client` and `V4Client`.

agentops/helpers/objects.py
dot-agi: Agreed. This has to be removed from here because its creating unnecessary code files.

dot-agi (changes requested):
Left some comments for improving code quality.

I was not able to test this PR with the `voice_transcription.py` file due to receiving the `ApiServerException` exception.

Is there a way to test this in realtime? I have two more notebooks to test from the OpenAI Cookbook website.

agentops/client/api/types.py
tcdent: This just matches the return values from the API: https://github.com/AgentOps-AI/AgentOps.Next/pull/817/files#diff-42cc0411597729fa5ab7ca4d2d7a85cafad38ef2922e915a8ce6cca919eabf76R35

agentops/client/api/versions/v4.py
tcdent: Overridden method. 

We're not sending JSON and we need to provide the `Authorization` header, so this is different.

agentops/helpers/config.py
tcdent: You want me to refactor `Client` in this PR?

tcdent:
> Left some comments for improving code quality.
> 
> I was not able to test this PR with the `voice_transcription.py` file due to receiving the `ApiServerException` exception.
> 
> Is there a way to test this in realtime? I have two more notebooks to test from the OpenAI Cookbook website.

You need to run the backend API from #817. 

`AGENTOPS_API_ENDPOINT="http://localhost:8000"; uv run examples/agents-example/voice_transcription.py`

examples/agents-example/voice_transcription.py
tcdent: Moved.

agentops/helpers/config.py
teocns: The way to do it is to open a different PR and rebase `agents-voice` onto its branch, if you will

dot-agi (approved):
challenged

dot-agi (approved):

PR #914, opened 2025-04-09
Add system stats and imported libraries tracking

Dwij1704:
## 📥 Pull Request

**📘 Description**
Closes #913 
Add system stats and imported libraries tracking

- Introduced `get_system_stats` function to collect CPU and memory information.
- Added `get_imported_libraries` function to retrieve top-level imported libraries.
- Updated `setup_telemetry` to include system stats and imported libraries in resource attributes.
- Expanded `ResourceAttributes` class to define new attributes for system and library information.

Note: `"os.type": "linux"` is default behaviour of OTEL.

**🧪 Testing**
```
  "resource_attributes": {
    "ProjectId": "6952a1d4-8a47-4c8b-98c5-1d03813ae5ad",
    "agentops.project.id": "6952a1d4-8a47-4c8b-98c5-1d03813ae5ad",
    "cpu.count": "16",
    "cpu.percent": "11",
    "host.machine": "AMD64",
    "host.name": "fd7ce180eb77",
    "host.node": "DESKTOP-ADFQBKV",
    "host.os_release": "10",
    "host.processor": "Intel64 Family 6 Model 154 Stepping 3, GenuineIntel",
    "host.system": "Windows",
    "host.version": "10.0.26100",
    "imported_libraries": "[\"agentops\",\"anthropic\",\"haystack\",\"litellm\",\"agents\",\"opentelemetry\"]",
    "memory.available": "3540471808",
    "memory.percent": "79",
    "memory.total": "16849293312",
    "memory.used": "13308821504",
    "os.type": "linux",
    "service.name": "agentops"
  },
```

PR #915, opened 2025-04-09
Add session management during shutdown and improve session handling

### LINKED ISSUE #894: agentops.init() Not Creating Session Span by Default; start_session() Creates Sibling Span Without Context Propagation

Dwij1704:
## 📥 Pull Request

**📘 Description**
Closes #914
Add session management during shutdown and improve session handling

- Introduced a global atexit handler to automatically end active sessions on shutdown.
- Updated the Client class to register the active session globally.
- Enhanced the end_session function to clear the client active session reference appropriately.
- Improved error handling during session ending to ensure robustness.
- Added context initialization in telemetry setup for better validation.

### DIFF
--- a/agentops/client/client.py
+++ b/agentops/client/client.py
@@ -1,4 +1,5 @@
 from typing import List, Optional, Union
+import atexit
 
 from agentops.client.api import ApiClient
 from agentops.config import Config
@@ -8,6 +9,28 @@ from agentops.logging import logger
 from agentops.logging.config import configure_logging, intercept_opentelemetry_logging
 from agentops.sdk.core import TracingCore
 
+# Global registry for active session
+_active_session = None
+
+# Single atexit handler registered flag
+_atexit_registered = False
+
+def _end_active_session():
+    """Global handler to end the active session during shutdown"""
+    global _active_session
+    if _active_session is not None:
+        logger.debug("Auto-ending active session during shutdown")
+        try:
+            from agentops.legacy import end_session
+            end_session(_active_session)
+        except Exception as e:
+            logger.warning(f"Error ending active session during shutdown: {e}")
+            # Final fallback: try to end the span directly
+            try:
+                if hasattr(_active_session, 'span') and hasattr(_active_session.span, 'end'):
+                    _active_session.span.end()
+            except:
+                pass
 
 class Client:
     """Singleton client for AgentOps service"""
@@ -61,6 +84,12 @@ class Client:
 
         self.initialized = True
 
+        # Register a single global atexit handler for session management
+        global _atexit_registered
+        if not _atexit_registered:
+            atexit.register(_end_active_session)
+            _atexit_registered = True
+
         # Start a session if auto_start_session is True
         session = None
         if self.config.auto_start_session:
@@ -71,6 +100,10 @@ class Client:
                 session = start_session(tags=list(self.config.default_tags))
             else:
                 session = start_session()
+            
+            # Register this session globally
+            global _active_session
+            _active_session = session
         
         return session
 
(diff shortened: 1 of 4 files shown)


braelyn-ai:
image

1) two session urls printed
2) incorrect API key should not be an exception

braelyn-ai:
double print is actually expected. one at kickoff and the other at the end of the run. API key was an issue with local environment.

nice job @Dwij1704 :)

agentops/client/client.py
```diff
@@ -61,6 +84,12 @@ def init(self, **kwargs):
 
         self.initialized = True
 
+        # Register a single global atexit handler for session management
```
braelyn-ai, line 87: nit: pls `session` -> `trace`

agentops/sdk/decorators/utility.py
```diff
@@ -233,18 +228,25 @@ def _record_entity_output(span: trace.Span, result: Any) -> None:
...
-    context_api.detach(token)
-    
-    # Debug info after detaching
-    final_context = _get_current_span_info()
-    logger.debug(f"[DEBUG] AFTER detach - Final context: {final_context}")
+    # Try to flush span processors
+    try:
+        from opentelemetry.trace import get_tracer_provider
+        tracer_provider = get_tracer_provider()
+        if hasattr(tracer_provider, "force_flush"):
```
braelyn-ai: please comment why it would or would not have force_flush

is there a better way to do this conditional than checking attributes?

braelyn-ai (approved):
this definitively limits us to one trace at a time. meaning we cannot track multiple agents running at once with the same agentops client.

this may need to be refactored for that in the future if/when that functionality is required

agentops/client/client.py
braelyn-ai: made a ticket for this in #916

braelyn-ai:
verified it works with manual testing connected to prod

agentops/sdk/decorators/utility.py
Dwij1704: Updated the docstring for the _finalize_span function to provide detailed explanations of its purpose, use cases, and arguments. Improved error handling by explicitly catching AttributeError when attempting to call force_flush on the tracer provider instead of condition for checking attributes.

braelyn-ai:
Is this ready to merge @Dwij1704 ?

Dwij1704:
> Is this ready to merge @Dwij1704 ?

Yes!

PR #919, opened 2025-04-11
bad api key doesnt block

braelyn-ai:
image

Allows the agent to run without AgentOps if the API key is invalid


Dwij1704:
We have two ways a user can use agentops
```
agentops.init()
```
or
```
agentops.init(auto_start_session=False)
agentops.start_session()
```

The fix doesn't work for the second one
image

Dwij1704:
> We have two ways a user can use agentops
> 
> ```
> agentops.init()
> ```
> 
> or
> 
> ```
> agentops.init(auto_start_session=False)
> agentops.start_session()
> ```
> 
> The fix doesn't work for the second one image

Fixed the same

Dwij1704 (approved):

PR #928, opened 2025-04-18
S3 logs

### LINKED ISSUE #878: Instrument `print`
captures all stdout

braelyn-ai:
All logs go to file, then at the end of a trace, the file is uploaded with your API key for storage in S3


Dwij1704:
The `agentops/common` directory is currently being used as a shared space for various instrumentation-related functions and utilities. However, logging seems to function as a separate instrumentor in itself. I suggest moving log instrumentation to `/agentops/logging` cause we might move instrumentation out of agentops for better maintenance, It makes sense to manage and update instrumentation packages independently rather than bumping the sdk itself every time an instrumentation needs update.

agentops/instrumentation/common/logs.py
```diff
@@ -0,0 +1,86 @@
...
+    """
+    ~Monkeypatches~ *Instruments the built-in print function and configures logging to also log to a file.
+    Preserves existing logging configuration and console output behavior.
+    """
+    log_file = os.path.join(os.getcwd(), LOGFILE_NAME)
+    
+    file_logger = logging.getLogger('agentops_file_logger')
+    file_logger.setLevel(logging.DEBUG)
+    
+    file_handler = logging.FileHandler(log_file)
```
tcdent: rather than creating a temp file could we just store the data to a variable?

we have to read the entire contents into memory for the upload anyway

braelyn-ai:
> The `agentops/common` directory is currently being used as a shared space for various instrumentation-related functions and utilities. However, logging seems to function as a separate instrumentor in itself. I suggest moving log instrumentation to `/agentops/logging` cause we might move instrumentation out of agentops for better maintenance, It makes sense to manage and update instrumentation packages independently rather than bumping the sdk itself every time an instrumentation needs update.

I totally understand this sentiment. We're instrumenting print and other loggers in this function so to me it felt like it belonged better in instrumentation, but I see what you're saying with instrumentation packages. i can move it :)

agentops/instrumentation/common/logs.py
braelyn-ai: that's a very fair point. I figured in case the run crashes but we dump on exception anyway

braelyn-ai: alex sed no

tcdent (approved):
LGTM

PR #976, opened 2025-05-13
Agent decorator fix

### LINKED ISSUE #968: [Bug]: @agent decorator doesn't work + not helpful
### Contact Details

_No response_

### 📦 Package Version

0.4.10

### 🎞️ Framework Version

`openai-1.78.0`

### 🔎 Describe the Bug

In the following script, we apply a `@agent` decorator to the GPTPlayer class. Agent spans don't show in the trace.

https://github.com/areibman/poke-env/blob/master/examples/gpt_player.py

Image

Additionally, we do not have the ability to add supplemental information to agents. Suppose I wanted to name this agent-- I can't! Something like `@agent(name="My agent")` would be fantastic.

This naming should apply to other decorators as well such as task, workflow, etc.

### 🤝 Contribution

fenilfaldu:
## 📥 Pull Request

**📘 Description**
Added async context manager support to handle span lifecycles in async operations, ensures proper cleanup and preventing premature span exports.

**🧪 Testing**
Verified span lifecycle management in async contexts and confirmed backward compatibility with existing sync initialization.


Dwij1704:
@fenilfaldu I have merged changes related to linting, please refactor to pass unit test :)

Dwij1704 (changes requested):
we can safely ditch __del__ here - the context managers in __init__ and __aexit__ have got our backs for both sync and async cleanup. It's like having two people doing the same job - might as well let the more reliable one handle it!

Dwij1704 (approved):
LGTM

PR #987, opened 2025-05-19
Lazy Loading of LLM Libraries to Prevent Duplicate Instrumentation

Dwij1704:
We've been facing issues with duplicate LLM instrumentation when multiple instrumentors (openai, crewai etc.) try to wrap the same LLM calls. This leads to redundant spans and potential conflicts in tracing.
Refactoring instrumentation module to be independent of each other is eating me up for the past week :)
## The Solution
Me, @the-praxs and @fenilfaldu were discussing approaches to fix tightly coupled behavior and make it independent of each other. Thanks to @the-praxs (🐐) who came up with this solution from V3, we've implemented a lazy loading approach for LLM libraries. Instead of eagerly importing and instrumenting all possible LLM providers and agentic frameworks, we now check sys.modules for direct imports and prioritize Agentic Libararis over providers to eliminate redundant spans.

## Why This is Better
- Prevents duplicate instrumentation of the same LLM calls
- Reduces overhead by only instrumenting what's actually used
- More reliable than depending on other instrumentors' LLM call spans
- Cleaner traces with single spans per LLM call

## Required Changes to Agentic Libraries
With this change, we need to update our agentic library integrations to record their own LLM spans:
- **CrewAI**: Need to modify to record spans for its internal LLM calls
- **AutoGen**: Update to track LLM interactions within agent conversations
- **Agents SDK**: Add span recording for direct LLM calls

This ensures that even when these frameworks make LLM calls internally, we'll have proper visibility into their operations without duplicate instrumentation.

## Current Limitations
There's one known limitation: if a provider/agentic framework is imported in a separate file and then imported into the main file, we miss it. For example:

```python
# llm_caller.py (helper file)
from anthropic import Anthropic  # This import might be missed

def call_anthropic():
    client = Anthropic()
    return client.messages.create(...)

# main.py
from llm_caller import call_anthropic  # We detect this import but miss the underlying Anthropic import
```
Possible fix is to scan imported modules' source files for additional imports

## Testing
I've done some testing with this approach and found it to be more reliable than our previous method. It successfully prevents duplicate spans but we still need to update crewai, autogen and agents instrumentations.

### DIFF
--- a/agentops/instrumentation/__init__.py
+++ b/agentops/instrumentation/__init__.py
@@ -149,28 +149,6 @@ def _check_existing_imports():
                 _instrumenting_packages.discard(root)
 
 
-def _start_monitoring_internal():
-    """Start monitoring imports and check already imported packages."""
-    builtins.__import__ = _import_monitor
-    _check_existing_imports()
-
-
-def _stop_monitoring_internal():
-    """Stop monitoring imports and restore the original import function."""
-    builtins.__import__ = _original_builtins_import
-
-
-def _uninstrument_all_internal():
-    """Stop monitoring and uninstrument all packages."""
-    global _active_instrumentors, _has_agentic_library
-    _stop_monitoring_internal()
-    for instrumentor in _active_instrumentors:
-        instrumentor.uninstrument()
-        logger.debug(f"Uninstrumented {instrumentor.__class__.__name__}")
-    _active_instrumentors = []
-    _has_agentic_library = False
-
-
 # Define the structure for instrumentor configurations
 class InstrumentorConfig(TypedDict):
     module_name: str
@@ -276,12 +254,19 @@ def instrument_all():
     """Start monitoring and instrumenting packages if not already started."""
     # Check if active_instrumentors is empty, as a proxy for not started.
     if not _active_instrumentors:
-        _start_monitoring_internal()
+        builtins.__import__ = _import_monitor
+        _check_existing_imports()
 
 
 def uninstrument_all():
     """Stop monitoring and uninstrument all packages."""
-    _uninstrument_all_internal()
+    global _active_instrumentors, _has_agentic_library
+    builtins.__import__ = _original_builtins_import
+    for instrumentor in _active_instrumentors:
+        instrumentor.uninstrument()
+        logger.debug(f"Uninstrumented {instrumentor.__class__.__name__}")
+    _active_instrumentors = []
+    _has_agentic_library = False
 
 
 def get_active_libraries() -> set[str]:
--- a/agentops/instrumentation/__init__.py
+++ b/agentops/instrumentation/__init__.py
@@ -1,18 +1,185 @@
-from typing import Optional
+"""
+AgentOps Instrumentation Module
+
+This module provides automatic instrumentation for various LLM providers and agentic libraries.
+It works by monitoring Python imports and automatically instrumenting packages as they are imported.
+
+Key Features:
+- Automatic detection and instrumentation of LLM providers (OpenAI, Anthropic, etc.)
+- Support for agentic libraries (CrewAI, AutoGen, etc.)
+- Version-aware instrumentation (only activates for supported versions)
+- Smart handling of provider vs agentic library conflicts
+- Non-intrusive monitoring using Python's import system
+"""
+
+from typing import Optional, Set
 from types import ModuleType
 from dataclasses import dataclass
 import importlib
 import sys
 from importlib.metadata import version
 from packaging.version import Version, parse
+import builtins
 
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
 
 from agentops.logging import logger
 from agentops.sdk.core import TracingCore
 
-_active_instrumentors: list[BaseInstrumentor] = []
 
+class InstrumentationManager:
+    """
+    Manages the instrumentation state and provides methods for monitoring and instrumenting packages.
+    This is implemented as a singleton to maintain consistent state across the application.
+    """
+
+    def __init__(self):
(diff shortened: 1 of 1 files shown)


Dwij1704:
Import Hooking Magic:
Hook into Python's import system, when someone imports an LLM provider or agentic library, we automatically detect it and set up the instrumentation.

We handle the tricky case of provider vs agentic library conflicts. For example, if someone's using CrewAI (which uses litellm, which uses OpenAI under the hood), we automatically uninstrument the direct OpenAI provider to avoid double-counting.

The implementation uses a singleton `InstrumentationManager` that:
- Monitors Python imports using a custom import hook
- Maintains a list of active instrumentors
- Handles the provider vs agentic library dance

@the-praxs @tcdent @areibman Thoughts? checkout the PR, I have made changes with this approach and it works!

dot-agi:
The dance of instrumentation is a tedious one.

But I will review thy PR. Time for some Easter eggs in the codebase :)

tcdent:
Does this now incorporate a fix for modules imported in sub-modules? Super common use case to have a project span multiple modules.

agentops/instrumentation/__init__.py
```diff
@@ -1,140 +1,296 @@
...
 
 from opentelemetry.instrumentation.instrumentor import BaseInstrumentor  # type: ignore
 
 from agentops.logging import logger
 from agentops.sdk.core import TracingCore
 
 
-# references to all active instrumentors
-_active_instrumentors: list[BaseInstrumentor] = []
+class InstrumentationManager:
```
tcdent: instead of a singleton can we just make this module-level functions?

agentops/instrumentation/__init__.py
```diff
@@ -1,140 +1,296 @@
...
+        self.stop_monitoring()
+        for instrumentor in self._active_instrumentors:
+            instrumentor.uninstrument()
+            logger.debug(f"Uninstrumented {instrumentor.__class__.__name__}")
+        self._active_instrumentors = []
+        self._has_agentic_library = False
+
+
+# Configuration for supported LLM providers
+PROVIDERS = {
```
tcdent: Can we use a `TypedDict` instead of bare dicts so these params are well documented?

tcdent (commented):
Just left a couple comments on code style.

Dwij1704:
> Does this now incorporate a fix for modules imported in sub-modules? Super common use case to have a project span multiple modules.

Yes it does, at first it checks using sys.modules to instrument existing direct imports, then relies on import hook to instrument indirect imports

Dwij1704:
@tcdent Try this out

#Example.py
```
load_dotenv()

import os
from openai import OpenAI
import agentops

agentops.init(os.getenv("AGENTOPS_API_KEY"))
from llm_caller import generate_content
client = OpenAI(api_key=os.getenv("OPENAI_API_KEY"))

print("Hello, world!")
res1 = client.chat.completions.create(
    model="gpt-4o-mini",
    messages=[{"role": "user", "content": "Hello, world!"}]
)
print(res1.choices[0].message.content)


response = generate_content("Hello, world!")
print(response)
```

#llm_caller.py
```
from anthropic import Anthropic
import os
load_dotenv()
client = Anthropic(api_key=os.getenv("ANTHROPIC_API_KEY"))
def generate_content(prompt):
    # Make a completion request - AgentOps will track it automatically
    message = client.messages.create(
        model="claude-3-opus-20240229",
        max_tokens=1000,
        messages=[
            {"role": "user", "content": prompt}
        ]
    )

    return message
```

agentops/instrumentation/__init__.py
Dwij1704: IMO the complexity involved (multiple state variables, several interacting methods) is substantial enough that a class helps keep things well-structured. Converting to module-level functions and globals might make the code feel more fragmented and the management of state less explicit.  I would advise against this particular refactoring unless there's a very strong reason related to specific constraints or a broader architectural goal not immediately apparent.

agentops/instrumentation/__init__.py
Dwij1704: Yeah thats better, pushing it right now

agentops/instrumentation/__init__.py
tcdent: entertain me for a sec just for the sake of growth:

we lose access to `_active_instrumentors` (unless we access it at `_manager._active_instrumentors`, which is just another layer of abstraction and supports my point). 

`_original_import` is really just a reference to the previous function and has nothing to do with state. only two methods need to know about it (`start_monitoring`, `stop_monitoring`)

I didn't read in too deep, but it seems like `_instrumenting_packages` could just reference  `_active_instrumentors` instead?

`_has_agentic_library` is just a flag and could be refactored to iterate through `_active_instrumentors` looking for `any(i.is_agentic for i in _active_instrumentors)`. better to have all of the instrumentor configs in one var and then use some logic (like a boolean flag) to separate them rather than keeping them in separate lists. even better if the instrumentor itself had control of the way it's perceived, but this global config array is fine for now since we don't have a concept of config inside instrumentors.

`instrument_all` and `uninstrument_all` just wrap the manager so they could be refactored to contain the actual logic; this is one of the big flags that brought up this note in the first place. 

you're feeling icky about having global state, but a singleton doesn't change that, it just makes it somewhat contained. IMO this is a pattern better reserved for languages that only have classes as primary logic structures, not a language like python that has module-level variables for this explicit use case.

Dwij1704: My silicon-based brain was perhaps a bit too attached to the old class structure.
Making those changes now :)

dot-agi (approved):

tcdent (approved):

Issue #988, opened 2025-05-19
[Feature]: Ability to rename root span names


areibman:
### 💡 Feature Description and Proposed Solution

<img width="309" alt="Image" src="https://github.com/user-attachments/assets/d3c00104-4f78-447a-804b-49a50d0d5dde" />

User was frustrated that everything is called "Session". Give a way in the SDK to rename the root span name

### 🤝 Contribution

- [ ] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the AgentOps team to handle this update.

Dwij1704:
Closed by #999

PR #990, opened 2025-05-19
Crewai root span fix

### LINKED ISSUE #960: [bug] No tags in crewai root span

fenilfaldu:
## 📥 Pull Request

**📘 Description**
Added support for customizing task span names using default tags and updated task span creation to use tag-based naming when available also maintained backward compatibility with default "crewai.task" name if no default-tags were passed while Initialization.

**🧪 Testing**
Tested the `mardown_validator.ipynb` runbook.

Screenshot 2025-05-20 at 5 05 25 AM

### DIFF
--- a/agentops/instrumentation/crewai/instrumentation.py
+++ b/agentops/instrumentation/crewai/instrumentation.py
@@ -14,6 +14,7 @@ from opentelemetry.sdk.resources import SERVICE_NAME, TELEMETRY_SDK_NAME, DEPLOY
 from agentops.instrumentation.crewai.version import __version__
 from agentops.semconv import SpanAttributes, AgentOpsSpanKindValues, Meters, ToolAttributes, MessageAttributes
 from .crewai_span_attributes import CrewAISpanAttributes, set_span_attribute
+from agentops import get_client
 
 
 # Initialize logger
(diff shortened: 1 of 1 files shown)


areibman:
Helpful! Can you also update the documentation to tell users how to provide this info?

Does this mean we could do `agentops.init(session_name="my session")` ?

fenilfaldu:
I would say we won't be able to do `agentops.init(session_name="my session")` because this tag can only be used when session is created and in cewai instrumentation `task` is the root span not the `session`.

Dwij1704 (commented):
@fenilfaldu You might have misunderstood this, the issue is that we don't set tags that we get from init in case of crewai, We need a mechanism to set list of tags in span attributes of root span (crewai.workflow etc). This can be done by setting it in the crewai instrumentor itself where the root span is first started.

The naming modifications you did is redundant for this issue, pick up https://github.com/AgentOps-AI/agentops/issues/988 after you are done with this.

I recommend closing this PR and raising a new one that only has changes related to tags in crewai and another PR with changes regarding setting span name for root spans that cover both sessions span and crewai specific root span types.

fenilfaldu:
Got it! closing this pr and raised the new one :)

PR #1030, opened 2025-06-02
Some coupling improvements. 

tcdent:
- [x] `make_span` and `finalize_span` were already tightly coupled to `TracingCore` but were living in another module. Make them part of core. 
- [x] `TracingCore` singleton convention was excessive; a single global instance is already thread-safe. 
- [x] `agentops.core` should be decoupled from semantc conventions. Moved span/resource attribute construction to a new file. 
- [ ] `agentops.helpers.system` probably has unused code we should clean up.


agentops/sdk/decorators/__init__.py
```diff
@@ -16,10 +16,11 @@
 operation_decorator = create_entity_decorator(SpanKind.OPERATION)
 workflow = create_entity_decorator(SpanKind.WORKFLOW)
 trace = create_entity_decorator(SpanKind.SESSION)
-session = create_entity_decorator(SpanKind.SESSION)
+# session = create_entity_decorator(SpanKind.SESSION)
```
braelyn-ai: comment or just delete?

braelyn-ai: or a deprecation warning on use

braelyn-ai (approved):

agentops/sdk/decorators/__init__.py
tcdent: `session` gets redefined as a function def just below this, so it was throwing an error in type checking. 

@Dwij1704 any input on this? can we just remove the one above?

Dwij1704: True, we can just remove it, We already have deprecation warning for it on use

braelyn-ai: yoink 'n merge

PR #1113, opened 2025-06-27
Concurrent fix

fenilfaldu:
## 📥 Pull Request

**📘 Description**
Previously, concurrent.futures was being instrumented globally whenever any provider or agentic library was imported. This caused unnecessary overhead and potential conflicts as the ThreadPoolExecutor instrumentation was applied universally, even when not needed.

This PR introduces a dependency-based instrumentation system that restricts concurrent.futures instrumentation to only activate when specific packages that require it are being used.

- Added UTILITY_DEPENDENCIES mapping defines which packages require which utility instrumentors.
- Currently maps mem0 to require concurrent.futures instrumentation.

**🧪 Testing**
[Example](https://gist.github.com/fenilfaldu/caa7f9b7b99708bbb560e2a8c97438c8)


Dwij1704 (approved):
GG

PR #1147, opened 2025-07-14
Prevent errors when instrumenting an agent using MCPToolset

ted-tai:
## 📥 Pull Request

**📘 Description**
Not all tools come with a name and description. This change prevents errors when an ADK Agent has a MCPToolset in its tools[] array and likely fixes other errors as well by checking a tool name and description exists before accessing them.

**🧪 Testing**
After the changes, I can use the MCPToolset without errors.


Dwij1704 (approved):

Dwij1704:
LGTM, Thanks @ted-tai

Issue #1025, opened 2025-05-30
[Feature]: Create Error Span Kind


areibman:
### 💡 Feature Description and Proposed Solution

Create an easy way to create and log errors as spans. This could be potentially BOTH a decorator and a standalone function. For example:

```
from agentops.sdk.decorators import error

@error
def log_error():
    ...

error()

```

OR

```
agentops.error(name="...", message="...")
```

### 🤝 Contribution

- [ ] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the AgentOps team to handle this update.

Br1an67:
I'd like to work on this. Added ERROR span kind as both a decorator (`@error`) and a standalone function (`agentops.error(name, message)`) that creates a one-shot span with error status.

PR: #1283

Issue #1024, opened 2025-05-30
[Feature]: Create arbitrary log spans


areibman:
### 💡 Feature Description and Proposed Solution

Langfuse has the functionality that lets users create log-style spans. They look something like this:
```
self.trace.event(name="xml_tool_call_limit_reached", level="DEFAULT", status_message=(
                                f"XML tool call limit reached - not yielding more content chunks"))
```

This is convenient for just showing where things might've gone wrong in a trace.

It would be great to have 2 functions:

* `agentops.log(...)`
* `agentops.error(...)`

### 🤝 Contribution

- [ ] Yes, I'd be happy to submit a pull request with these changes.
- [ ] I need some guidance on how to contribute.
- [ ] I'd prefer the AgentOps team to handle this update.

dot-agi:
I have create a `TraceState` enum in #1015 which has 3 `StatusCode` from OpenTelemetry mapped to strings.

I think having `agentops.set_state(<state in str or agentops.<STATE>>)` would be a better UX.

Br1an67:
I'd like to work on this. Added both `agentops.log(name, message, level)` and `agentops.error(name, message)` as standalone one-shot span functions, plus `@log`/`@error` decorators.

PR: #1283

PR #1283, opened 2026-03-01
feat: add error and log span kinds

Br1an67:
## 📥 Pull Request

**📘 Description**

Add `ERROR` and `LOG` span kinds to the SDK, addressing both #1025 and #1024.

Each span kind is available in two forms:

**Decorators** — wrap functions to trace them as error/log spans:
```python
from agentops.sdk.decorators import error, log

@error
def handle_failure():
    ...

@log(name="custom_log")
def log_event():
    ...
```

**Standalone functions** — create one-shot event spans:
```python
import agentops

agentops.error(name="validation_failed", message="Invalid input format")
agentops.log(name="checkpoint", message="Processing complete", level="INFO")
```

Changes:
- `agentops/semconv/span_kinds.py`: Add `ERROR` and `LOG` to `AgentOpsSpanKindValues` enum and `SpanKind` class
- `agentops/sdk/decorators/__init__.py`: Create `error`/`log` decorators via `create_entity_decorator` factory
- `agentops/__init__.py`: Add standalone `error()` and `log()` functions using `_create_as_current_span`, update exports

**🧪 Testing**

Added 14 unit tests in `tests/unit/sdk/test_error_log_spans.py` covering:
- Error decorator: basic span creation, custom name, async support
- Log decorator: basic span creation, custom name
- Standalone `error()`: span creation, error status code, extra attributes, default name
- Standalone `log()`: span creation, log level, default INFO level, extra attributes, default name

All 164 tests pass (including existing tests). Ruff lint clean.

### DIFF
--- a/agentops/sdk/decorators/__init__.py
+++ b/agentops/sdk/decorators/__init__.py
@@ -17,6 +17,8 @@ tool = create_entity_decorator(SpanKind.TOOL)
 operation = task
 guardrail = create_entity_decorator(SpanKind.GUARDRAIL)
 track_endpoint = create_entity_decorator(SpanKind.HTTP)
+error = create_entity_decorator(SpanKind.ERROR)
+log = create_entity_decorator(SpanKind.LOG)
 
 
 # For backward compatibility: @session decorator calls @trace decorator
@@ -48,4 +50,6 @@ __all__ = [
     "tool",
     "guardrail",
     "track_endpoint",
+    "error",
+    "log",
 ]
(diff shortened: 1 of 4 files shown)


agentops/sdk/decorators/__init__.py
```diff
@@ -17,6 +17,8 @@
 operation = task
 guardrail = create_entity_decorator(SpanKind.GUARDRAIL)
 track_endpoint = create_entity_decorator(SpanKind.HTTP)
+error = create_entity_decorator(SpanKind.ERROR)
```
RichardoMrMu, line 20: [P1] Set OTel error status for successful @error calls

The new @error decorator only selects the custom agentops.span.kind=error attribute through the generic entity factory. When the decorated function returns normally—the exact behavior exercised by all three new decorator tests—the factory never calls span.set_status(StatusCode.ERROR), so the exported OTel span remains UNSET and is omitted from standard status.code=ERROR queries and alerts. This also makes @error disagree with the new agentops.error() helper, which does set ERROR. Please make the error decorator set ERROR for its normal-return path as well and assert the OTel status in the sync and async regressions.

PR #1431, opened 2026-07-25
fix(agno): clean up async streams on iteration errors

### LINKED ISSUE #1430: Agno async stream wrapper leaks span/context on post-yield errors (span never ended, no ERROR status)
**Describe the bug**

`AsyncStreamingResultWrapper.__anext__` in the Agno instrumentor only handles `StopAsyncIteration`. If the wrapped async stream raises any other exception after yielding at least one event, the span is never ended and no error is recorded, and a stale entry is left in the context manager.

https://github.com/AgentOps-AI/agentops/blob/f8e907b92dabe47232978023fdcb01e2a7d4b752/agentops/instrumentation/agentic/agno/instrumentor.py#L133-L147

The outer agent wrapper sets `span.set_status(OK)` before returning the wrapper and returns it, so its own `try/except` cannot catch errors raised later during consumer iteration. `__anext__` catches only `StopAsyncIteration`; its `finally` merely detaches the context token. Any other exception mid-stream therefore skips `span.set_status(ERROR)`, `span.record_exception`, `span.end()`, and `remove_context`.

Net effect: an errored async Agent/Team stream produces a span that is never ended (never exported), with no error recorded, plus a stale `_contexts` entry. For an observability SDK this is a telemetry-correctness defect on a realistic path (LLM streams failing mid-iteration).

**Reproduction**

1. Instrument an Agno `Agent.arun(..., stream=True)` (or `Team.arun`) whose underlying async generator raises after yielding at least one event.
2. Observe that no span is exported for the failed stream, no exception is recorded on it, and a stale entry remains in the context manager.

**Suggested fix**

In `AsyncStreamingResultWrapper.__anext__`, add an `except Exception as e:` that sets `span.set_status(ERROR)`, `span.record_exception(e)`, `span.end()`, and `remove_context(...)` (guarded so it runs once), then re-raises, mirroring the sync path and the outer wrapper's handler.

**Environment:** current `main` (HEAD `f8e907b` at time of report). Path introduced by #1265.

---

_Found while testing [Ito](https://ito.ai), an automated code-review tool, against recently-merged PRs. It's free for open source. Sharing this because it looked like a real bug worth fixing, not to sell anything: https://app.ito.ai/share/8c8baa02-5ba9-416d-b5bb-a7d2e8792c35?tab=details_

KXHXK:
## Description

- handle exceptions raised after an Agno async stream has started yielding
- mark the span as `ERROR`, record the original exception, end the span, and remove its stored context exactly once
- re-raise the original stream exception to preserve caller behavior
- add an offline regression test for the post-yield failure path

Without this cleanup, failed async Agent/Team streams leave an unended span and a stale entry in `StreamingContextManager`.

Closes #1430

## Testing

- `uv run ruff check agentops/instrumentation/agentic/agno/instrumentor.py tests/unit/instrumentation/test_agno_instrumentor.py`
- `uv run ruff format --check agentops/instrumentation/agentic/agno/instrumentor.py tests/unit/instrumentation/test_agno_instrumentor.py`
- `uvx pre-commit run --files agentops/instrumentation/agentic/agno/instrumentor.py tests/unit/instrumentation/test_agno_instrumentor.py`
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-sync pytest -p pytest_asyncio.plugin -p pytest_mock tests/unit/instrumentation -q` (`139 passed`)


agentops/instrumentation/agentic/agno/instrumentor.py
```diff
@@ -143,6 +143,14 @@ async def __anext__(self):
                 self.span.end()
                 self.streaming_context_manager.remove_context(self.agent_id)
             raise
+        except Exception as e:
```
RichardoMrMu, line 155: **[P2] Clean up cancelled async iterations**

`asyncio.CancelledError` directly subclasses `BaseException` on every supported Python version, so a timeout or task cancellation raised by `await self.original_result.__anext__()` bypasses this new `except Exception` branch. The `finally` block only detaches the OTel context; `_consumed` stays false, the span is never ended, and the `StreamingContextManager` entry remains. This leaves the same telemetry/resource leak this PR fixes for `RuntimeError` on a normal cancellation path. Please perform the end/remove cleanup for `CancelledError` and then re-raise it, with a cancelled-task regression.

KXHXK: Addressed in `57d256d`. `AsyncStreamingResultWrapper` now handles `asyncio.CancelledError` explicitly, marks the span ERROR, records and re-raises the cancellation, ends the span, and removes the stored context exactly once. The new regression repeats `__anext__()` after cancellation to verify cleanup stays idempotent. The changed files pass `py_compile`, and an isolated module-level cancellation harness passes.

RichardoMrMu: Rechecked on 57d256d. The explicit asyncio.CancelledError branch now performs guarded status/exception/span-end/context-removal cleanup before a bare re-raise, while finally detaches the OTel context. The regression covers post-yield cancellation and verifies cleanup stays idempotent. This closes my finding—thanks.

PR #1435, opened 2026-07-25
fix: preserve stream error status

### LINKED ISSUE #1434: Stream finalization overwrites iteration errors with OK status
## Description

The shared sync and async stream wrappers overwrite an iteration failure with a successful span status during finalization.

Both wrappers catch an iteration exception, set the span to `ERROR`, record the exception, and re-raise it. Their `finally` blocks then call `BaseStreamWrapper._finalize()`, which unconditionally sets the same span to `OK` before ending it.

The final exported status can therefore report a failed provider stream as successful even though the original exception is preserved for the caller.

## Minimal reproduction

```python
from unittest.mock import Mock

from agentops.instrumentation.common.streaming import SyncStreamWrapper


def failing_stream():
    yield chunk
    raise ValueError(boom)


span = Mock()
try:
    list(SyncStreamWrapper(failing_stream(), span, lambda chunk: chunk))
except ValueError:
    pass

print([call.args[0].status_code.name for call in span.set_status.call_args_list])
```

Current output on `main`:

```text
['ERROR', 'OK']
```

The async wrapper follows the same control flow.

## Expected behavior

Successful streams should end with `OK`. Failed streams should record the original exception and end with `ERROR`, without a later `OK` status update.

## Suggested fix

Let `_finalize()` know whether iteration completed successfully, or move the success status assignment out of unconditional finalization. Add sync and async regression assertions for the final status.

KXHXK:
## Description

Fixes #1434.

The common synchronous and asynchronous stream wrappers currently mark a span as `ERROR` when iteration raises, then overwrite that status with `OK` during finalization. This change keeps final attribute collection and span closure intact while skipping the success status after an iteration failure.

Both wrapper variants now have regression coverage asserting that the emitted status remains `ERROR`.

## Testing

- `tests/unit/instrumentation/common`: 56 passed
- `ruff check`: passed
- `ruff format --check`: passed
- `git diff --check`: passed
- secret-pattern scan of changed files: passed

Direct `mypy agentops/instrumentation/common/streaming.py` reports the pre-existing unannotated `accumulated_content` list on line 36; the same line is present on `upstream/main` and is outside this patch.


agentops/instrumentation/common/streaming.py
```diff
@@ -100,32 +105,36 @@ class SyncStreamWrapper(BaseStreamWrapper):
...
                 self._process_chunk(chunk)
                 yield chunk
         except Exception as e:
+            failed = True
             self.span.set_status(Status(StatusCode.ERROR, str(e)))
             self.span.record_exception(e)
             raise
         finally:
-            self._finalize()
+            self._finalize(set_success_status=not failed)
```
RichardoMrMu, line 143: **[P2] Do not mark cancelled streams successful**

On every supported Python version, `asyncio.CancelledError` directly subclasses `BaseException`, so a cancellation raised during this `async for` bypasses `except Exception` and leaves `failed = False`. This `finally` path then calls `_finalize(set_success_status=True)`, which emits `OK` before ending the span even though the consumer task was cancelled. Timeouts and shutdown cancellation are therefore still reported as successes—the status inversion this PR is intended to prevent. Please give `CancelledError` an explicit non-success cleanup path, re-raise it, and add a cancelled-iterator regression.

KXHXK: Addressed in `a11e853`. `AsyncStreamWrapper` now catches `asyncio.CancelledError` explicitly, records an ERROR status and the cancellation, finalizes without writing a later OK status, and re-raises the same cancellation. Added a cancelled-iterator regression asserting the status sequence is exactly `[ERROR]`, the original exception is recorded, and the span ends once. The changed files pass `py_compile`, and an isolated module-level cancellation harness passes.

RichardoMrMu: Rechecked on a11e853. Cancellation now records only ERROR, re-raises the same asyncio.CancelledError, and finalizes/ends once. The regression asserts that external status and cleanup contract directly. This closes my finding.

PR #1437, opened 2026-07-27
fix(openai): preserve response output content index

### LINKED ISSUE #1436: OpenAI Responses message content uses the wrong completion index
## Description

`get_response_output_message_attributes()` uses the index within `message.content` when extracting `ResponseOutputText`, instead of the message's index within `response.output`.

This mis-associates message text whenever another output item (commonly a reasoning item) precedes the message. The message metadata is emitted under `gen_ai.completion.1.*`, while its content is emitted under `gen_ai.completion.0.content` and can overwrite attributes belonging to the reasoning item.

Reproduced on current `main` at `f8e907b92dabe47232978023fdcb01e2a7d4b752`.

## Minimal reproduction

```python
from openai.types.responses import (
    ResponseOutputMessage,
    ResponseOutputText,
    ResponseReasoningItem,
)

from agentops.instrumentation.providers.openai.attributes.response import (
    get_response_output_attributes,
)

reasoning = ResponseReasoningItem(
    id="reasoning_1", summary=[], type="reasoning", status="completed"
)
message = ResponseOutputMessage(
    id="msg_1",
    content=[ResponseOutputText(annotations=[], text="answer", type="output_text")],
    role="assistant",
    status="completed",
    type="message",
)

print(get_response_output_attributes([reasoning, message]))
```

Relevant current output:

```text
gen_ai.completion.0.id = reasoning_1
gen_ai.completion.0.content = answer
gen_ai.completion.1.id = msg_1
# gen_ai.completion.1.content is missing
```

## Expected behavior

The text content should use the parent output message's completion index:

```text
gen_ai.completion.0.id = reasoning_1
gen_ai.completion.1.id = msg_1
gen_ai.completion.1.content = answer
```

## Suggested fix

Pass the outer `response.output` index through to `get_response_output_text_attributes()` instead of replacing it with the inner content-list index. Add a regression using actual OpenAI response types with a reasoning item before the message.

I can submit a focused fix with regression coverage.

yaodong-shen:
## 📥 Pull Request

**📘 Description**

Fixes #1436.

OpenAI Responses can return a reasoning output before the assistant message. The message metadata correctly used its index in `response.output`, but nested `ResponseOutputText` attributes used the index within `message.content`. This placed message text under the preceding reasoning completion and left the actual message completion without content.

This change preserves the parent output index while extracting nested message text. The regression test uses the real OpenAI response models and verifies that a reasoning item at index 0 cannot absorb message content from index 1.

**🧪 Testing**

- `.venv/bin/python -m pytest tests/unit/instrumentation/openai_core/test_response_attributes.py -q -o log_cli=false` — 14 passed
- `.venv/bin/python -m pytest tests/unit -q -o log_cli=false` — 489 passed, 1 skipped
- targeted Ruff check and format check — passed
- project pre-commit hooks on both changed files — passed
- `git diff --check` — passed


agentops/instrumentation/providers/openai/attributes/response.py
```diff
@@ -462,9 +462,9 @@ def get_response_output_message_attributes(index: int, message: "ResponseOutputM
     attributes = _extract_attributes_from_mapping_with_index(message, RESPONSE_OUTPUT_MESSAGE_ATTRIBUTES, index)
 
     if message.content:
-        for i, content in enumerate(message.content):
+        for content in message.content:
             if isinstance(content, ResponseOutputText):
-                attributes.update(get_response_output_text_attributes(content, i))
+                attributes.update(get_response_output_text_attributes(content, index))
```
RichardoMrMu, line 467: [P2] Preserve every text block in a response message

A ResponseOutputMessage.content list can legally contain multiple ResponseOutputText blocks. Passing the same parent index for each block makes every helper call emit gen_ai.completion.{index}.content, so attributes.update() silently replaces all earlier text with the last block. The helper also restarts each block's annotation index at j=0, causing later annotations to overwrite earlier ones under the same completion. Please aggregate the text blocks in order and keep annotation indices unique across the whole message (with offsets consistent with the aggregated content), and add a regression containing two annotated output_text blocks.

Issue #1447, opened 2026-09-03
old semconv attribute names (prompt_tokens vs input_tokens) — cross-SDK aggregation gap


roy-tong:
A follow-up to #1445 (Anthropic cache tokens), from the same 30-project conformance audit.

**The observation**

AgentOps uses the older GenAI semconv attribute names in `agentops/semconv/span_attributes.py` (~L61-63):
```python
LLM_USAGE_PROMPT_TOKENS = "gen_ai.usage.prompt_tokens"
LLM_USAGE_COMPLETION_TOKENS = "gen_ai.usage.completion_tokens"
```

The current OTel GenAI semantic conventions use `gen_ai.usage.input_tokens` / `gen_ai.usage.output_tokens`. OpenLLMetry, pydantic-ai, and other instrumentation we audited have already migrated to the new names.

The file header (~L8-24) contains TODO comments acknowledging mapping issues.

**The consequence**

In an OTLP collector receiving spans from multiple SDKs:
- `sum(gen_ai.usage.input_tokens)` misses all AgentOps spans (they emit `prompt_tokens`)
- `sum(gen_ai.usage.prompt_tokens)` misses all OpenLLMetry spans (they emit `input_tokens`)
- Cross-SDK token aggregation requires knowing which attribute name each SDK uses — per-SDK knowledge that generic consumers shouldn't need

**The question**

Is there a plan to migrate to the current `gen_ai.usage.input_tokens` / `output_tokens` names, or to emit both during a transition? The old names still work within the AgentOps ecosystem, but any downstream tool consuming generic OTel GenAI attributes will silently undercount when AgentOps spans are mixed in.

Happy to send a two-SDK repro (one AgentOps span + one OpenLLMetry span in the same collector, asserting the aggregation gap).

Context: AgentMeasure conformance check; this is the AgentOps follow-up after the Anthropic cache-token finding in #1445.

ZxlDragonDoctor:
I'd like to take this. Plan: dual-emit current OTel GenAI usage names (`gen_ai.usage.input_tokens` / `output_tokens`) alongside the existing `prompt_tokens` / `completion_tokens` in `TokenUsage.to_attributes()`, with regression coverage, so cross-SDK collectors can sum either family during the transition. No call-site rewrite in this PR.

roy-tong:
[...]

PR #1450, opened 2026-09-09
fix(anthropic): emit prompt cache usage attributes

### LINKED ISSUE #1445: Anthropic prompt caching: cache tokens not emitted on the Anthropic paths — is session total_tokens systematically low?
A semantics question found while auditing token accounting across agent-telemetry SDKs.

**The observation**

On the Anthropic paths, usage extraction reads only `usage.input_tokens` / `usage.output_tokens` and computes `total = input + output`:

- non-streaming: `agentops/instrumentation/providers/anthropic/attributes/message.py` (~L319-332)
- streaming: `.../anthropic/stream_wrapper.py` (~L167-177 and ~L406-418)

Anthropic's `cache_read_input_tokens` / `cache_creation_input_tokens` are separate fields (not included in `input_tokens`), and I don't find them read anywhere on these paths — `LLM_USAGE_CACHE_READ_INPUT_TOKENS` / `CACHE_CREATION_INPUT_TOKENS` are defined in `agentops/semconv/span_attributes.py` but have no setter calls in the Anthropic instrumentation.

Meanwhile the OpenAI Responses path (`providers/openai/attributes/response.py`, ~L296-307) does emit `cached_tokens → cache_read` and `reasoning_tokens` — so the asymmetry is within one SDK.

**Why it matters**

The OTel GenAI semantic conventions for Anthropic (`semantic-conventions-genai`, `docs/gen-ai/anthropic.md`) are explicit: cache read/write tokens **MUST be added** to the native `input_tokens` to compute `gen_ai.usage.input_tokens`. If they're dropped instead, a prompt-caching Claude session reports `total_tokens` systematically below what the provider actually meters (cache reads are billed, just at a lower rate — they're not free tokens).

**The question**

Is this asymmetry intentional (e.g. a deliberate "native usage only" convention on the Anthropic path)? If not: for sessions with `cache_control` enabled, does any downstream token/cost figure derived from these spans undercount? Happy to send a minimal repro — same prompt called twice with caching on, asserting the cache attributes are absent and the total vs the Anthropic console's billed tokens diverge — or a small PR adding the cache fields per the semconv rule.

Context: I maintain AgentMeasure, a measurement-conformance project running a public 30-projects/30-days check of agent-telemetry semantics; this is the finding worth asking about directly. Not claiming a bug without your read on the intended convention.

jstar0:
## Summary

- Extract Anthropic prompt-cache read and creation token counts from non-streaming messages and raw streaming metadata.
- Keep the existing prompt, completion, and total attributes based on Anthropic native input/output values.
- Cover synchronous and asynchronous stream final-message paths with fixture-based regressions.

## Why

Anthropic reports cache-read and cache-creation tokens as fields separate from input_tokens. The instrumentation previously dropped those fields, making cache activity invisible to telemetry consumers. This change exposes both fields independently while preserving the existing native input/output total, so consumers can choose their accounting policy without double-counting cache tokens.

## Validation

- uv run pytest -q tests/unit/instrumentation/anthropic/test_attributes.py tests/unit/instrumentation/anthropic/test_stream_wrapper.py — 22 passed
- uv run ruff check on the four changed files — passed
- uv run ruff format --check on the four changed files — passed

Fixes #1445


roy-tong (approved):
Reviewed — the semantics are right and the docstring states them well: Anthropic's `input_tokens` excludes cached buckets, so exposing `cache_read_input_tokens` / `cache_creation_input_tokens` as independent facets without touching the native totals is the correct shape, and `is not None` correctly keeps measured zeros distinct from absent. Approving with one cross-provider consistency point worth settling here rather than later:

**`TOTAL_TOKENS = input + output` is a third grain, not a shared one.** On the OpenAI path (and in the dual-emit from #1453), `input_tokens` already includes cached tokens, so `total = input + output` is the inclusive sum. On this path it is not: Anthropic-exclusive input plus output omits both cache buckets. A collector summing `gen_ai.usage.total_tokens` across a mixed OpenAI+Anthropic workload is aggregating two different grains — the same cross-SDK hazard #1447/#1453 exist to close, one attribute earlier than anyone looks. Either:

- make this path's total inclusive too (`input + cache_read + cache_creation + output`, matching the OpenAI-side grain), or
- keep it native and document in the docstring that `total_tokens` here excludes cache buckets and consumers must add their own policy.

I'd lean inclusive for consistency, but either is defensible if the docstring names the basis. Non-blocking either way since the previous code had the same total.

One forward note (already on record in my #1453 review): these flat cache attribute names are the same legacy generation that will eventually need dual-emitting alongside the current nested `gen_ai.usage.cache_read.input_tokens` form — worth keeping in the same transition tracking.

